PR #21, opened 2017-12-16
Support passing arguments to scripts (fixes #18)

### LINKED ISSUE #18: No way to pass arguments to scripts run under monkeytype
I want to run

`myscript update`

but `monkeytype run myscript update` 

errors with:

`monkeytype: error: unrecognized arguments: update`

folz:
This turned out to be an easier PR than I expected. run_handler/run_path already use sys.argv directly; I just added a catchall argument to the argparser so it doesn't complain about additional tokens.

Doesn't look like there are tests for this in test_cli.py, but let me know if I should create a new one.

### DIFF
--- a/monkeytype/cli.py
+++ b/monkeytype/cli.py
@@ -195,6 +195,10 @@ def main(argv: List[str], stdout: IO, stderr: IO) -> int:
         'script_path',
         type=str,
         help="""Filesystem path to a Python script file to run under tracing""")
+    run_parser.add_argument(
+        'script_args',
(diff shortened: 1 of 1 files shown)


carljm:
This looks great, thank you! We don’t currently have tests covering argument parsing; wouldn’t be a bad idea to add some, but certainly not going to block this fix on it.

folz:
My pleasure, and thank you for releasing monkeytype! It's already made my work (and codebase) better.

PR #23, opened 2017-12-16
Generate annotations for NewType types (fixes #22)

### LINKED ISSUE #22: `NewType` generated into invalid type signatures
Given

```python
UserId = NewType('UserId', int)

def get_user(self, user_id: UserId) -> 'User':
```

MonkeyType generates

```python
def get_user(
    self,
    user_id: <function NewType.<locals>.new_type at 0x1090a52f0>
) -> 'User': ...
```

which then causes retype to fail with this error message:

```
error: /path/to/module.py: invalid syntax (<unknown>, line 18)
```

This happens even when `--exclude-unparsable-defaults` is passed.

I was able to come up with a solution to this, which I'll submit a PR for. However, it seems hacky to me and I'd appreciate feedback on it.

folz:
Per #22, monkeytype generates invalid type signatures for NewType types because they are valid mypy annotations but `new_type()` gets inspected as a function at runtime. This PR special-cases support for objects with `__supertype__` to return their `__name__` instead of their repr, based on https://github.com/python/cpython/blob/v3.6.3/Lib/typing.py#L2227

Since a comment says `render_annotation` was derived from `inspect.formatannotation`, I'll point out that this PR causes render_annotation to disagree with inspect.formatannotation. I think this is okay in theory since render_annotation is used for mypy types while formatannotation is used for runtime inspection, but I'd appreciate a second opinion. For what it's worth, this patch allowed me to use monkeytype against my NewType-using codebase successfully.

This seems important enough to warrant tests - if you think this PR is taking a good approach, I can write tests for it as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ htmlcov/
 doc/_build
 dist/
 build/
+.idea/
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,7 +5,7 @@ master
 ------
 
 * Fix passing args to script run with ``monkeytype run`` (#18; merge of #21).
-
+* Fix generated annotations for NewType types (#22; merge of #23).
 
 17.12.1
 -------
--- a/monkeytype/stubs.py
+++ b/monkeytype/stubs.py
@@ -319,6 +319,8 @@ def render_annotation(anno: Any) -> str:
         elem_type = _get_optional_elem(anno)
         rendered = render_annotation(elem_type)
         return 'Optional[' + rendered + ']'
+    elif hasattr(anno, '__supertype__'):
+        return anno.__name__
     elif getattr(anno, '__module__', None) == 'typing':
         return repr(anno).replace('typing.', '')
     elif anno is NoneType:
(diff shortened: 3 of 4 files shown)


carljm:
This approach looks fine to me; thanks for the report and fix!

Normally we try to automatically provide the right imports for annotations too, so `monkeytype apply` doesn't result in modules with missing imports. However, I don't think that's possible with the current runtime implementation of `NewType`; the calling module is not stored anywhere in the resulting function object, so there's no way we can introspect it. So unless we change that in `typing.py` we're stuck with the limitation of not being able to automatically add NewType imports.

It would be great to add tests for NewType in `test_stubs.py`; thanks!

carljm:
Er, never mind about generating imports; `NewType` annotations can only occur as existing annotations, never as traced types. So the import should always be present already.

folz:
Added a test for this and updated the changelog. I think you're right that this doesn't need to worry about imports - if the import isn't present, it wasn't valid to mypy in the first place.

carljm:
Awesome, thank you!

Issue #28, opened 2017-12-18
Custom module import behavior (support for Django)


folz:
Running monkeytype against a file with Django models or dependencies on Django apps gives me a series of errors.

First, just `monkeytype apply package.app.models`:

```
django.core.exceptions.ImproperlyConfigured: Requested setting DEFAULT_INDEX_TABLESPACE, but settings are not configured. You must either define the environment variable DJANGO_SETTINGS_MODULE or call settings.configure() before accessing settings.
```

Second, `DJANGO_SETTINGS_MODULE=config.settings.local monkeytype apply package.app.models`:

```
File "package/app/models.py"
django.core.exceptions.AppRegistryNotReady: Apps aren't loaded yet.
```

Third, adding

```python
import django
django.setup()
```

to the top of package.app.models:

```
File "some/other/models.py"
django.core.exceptions.AppRegistryNotReady: Apps aren't loaded yet.
```

What **has** worked for me is patching `monkeytype/cli.py` and `monkeytype/util.py` to add

```python
+   import django
+   django.setup()
    mod = importlib.import_module(module)
```

Obviously this isn't suitable as-is because it breaks monkeytype on non-Django codebases. However, it did give me the idea to have an `ModuleImporter` (name WIP) context manager that returns `importlib.import_module(module)` but may also perform pre-/post-import actions. It would work similarly to TypeRewriter: the default implementation could just `import_module`, but a DjangoModuleImporter could call `django.setup()`, and the specific ModuleImporter could be configured in settings.

Thoughts? Happy to submit a PR for this if you think it's the right direction.

carljm:
Ugh, modules that raise errors at import time :)

Yeah, I think we need some way to inject this kind of setup. I'm not sure about wrapping every call to `import_module`, though; `get_name_in_module` could be called many times in the same MonkeyType run, we don't really want or need to call `django.setup` every time, do we?

We could just add a config method called something like `cli_setup()` that gets called once at startup of the CLI?

folz:
Makes sense, yeah. Opened a PR based on your comment.

carljm:
Fixed with merge of #29.

lsh-0:
Using `cli_context` in the linked fix isn't working for me with django ([as per the docs](http://monkeytype.readthedocs.io/en/latest/configuration.html#monkeytype.config.Config.cli_context)) with this as my `monkeytype_config.py` file:

```
from monkeytype.config import DefaultConfig
from contextlib import contextmanager

[...]

PR #29, opened 2017-12-19
Add Config.cli_context() as a hook for custom CLI initialization and cleanup logic.

folz:
Addressing #28, this PR adds the `cli_context()` method to Config so the CLI can use it as a context manager when running its command.

The command name is passed so implementations can, e.g., not run `django.setup()` with `monkeytype run manage.py` (as that would be redundant) but still run it for `stub`/`apply`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,6 @@ Changelog
 master
 ------
 
-
 17.12.2
 -------
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,7 @@ Changelog
 master
 ------
 
-* Add setup/teardown CLI lifecycle hooks.
+* Add Config.cli_context() as a hook for custom CLI initialization and cleanup logic (#28; merge of #29).
 
 17.12.2
 -------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,8 @@ Changelog
 master
 ------
 
+* Add setup/teardown CLI lifecycle hooks.
+
 17.12.2
 -------
 
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -99,6 +99,32 @@ Subclassing ``Config`` or ``DefaultConfig``
 
     Defaults to ``False``.
 
+  .. method:: cli_setup(command: str) -> None
+
+    A lifecycle hook which gets called once, right after the CLI starts. Use it
+    to perform any initialization required by your codebase.
+
+    ``command`` is the command passed to the ``monkeytype`` cli: ``'run'``,
+    ``'apply'``, etc.
+
+    For example, if you run MonkeyType against a Django codebase, you can use
+    this hook to call::
+
+      import django
+      django.setup()
+
+    The default implementation of this method does nothing.
+
+  .. method:: cli_teardown(command: str) -> None
+
+    A lifecycle hook which gets called once, right before the CLI exits. Use it
+    to undo side effects from :meth:`cli_setup`, if applicable.
+
+    ``command`` is the command passed to the ``monkeytype`` cli: ``'run'``,
+    ``'apply'``, etc.
+
+    The default implementation of this method does nothing.
+
 .. class:: DefaultConfig()
 
   ``DefaultConfig`` is the config MonkeyType uses if you don't provide your own;
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -99,31 +99,32 @@ Subclassing ``Config`` or ``DefaultConfig``
 
     Defaults to ``False``.
 
-  .. method:: cli_setup(command: str) -> None
+  .. method:: cli_context(command: str) -> Iterator[None]
 
-    A lifecycle hook which gets called once, right after the CLI starts. Use it
-    to perform any initialization required by your codebase.
+    A context manager which wraps the execution of the CLI command. Use it to
+    perform any initialization or cleanup required by your codebase.
 
-    ``command`` is the command passed to the ``monkeytype`` cli: ``'run'``,
+    ``command`` is the command passed to the monkeytype cli: ``'run'``,
     ``'apply'``, etc.
+
+    Since :meth:`cli_context` is used as a context manager, any :class:`Config`
+    subclass that implements this method needs to decorate it with a context
+    decorator like :meth:`contextlib.contextmanager`, and needs to yield in
+    the body of the method exactly once. Code that comes before the yield will
+    run before the command starts, and code that comes afterward will run when
+    the command finishes but before the CLI exits.
 
     For example, if you run MonkeyType against a Django codebase, you can use
-    this hook to call::
+    this method to setup Django before the command runs::
 
-      import django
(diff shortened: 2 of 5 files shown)


CHANGES.rst
```diff
@@ -4,6 +4,7 @@ Changelog
 master
 ------
 
+* Add setup/teardown CLI lifecycle hooks.
```
carljm: Mention the GitHub issue number (and the PR number too, now that it exists.)

folz:
`cli_context` works for me. In addition to the documentation, I updated the CLI `main()` with friendly error messages for common context manager mistakes. If you think this makes the implementation too complicated I can take that back out.

doc/configuration.rst
```diff
@@ -99,6 +99,33 @@ Subclassing ``Config`` or ``DefaultConfig``
...
+    subclass that implements this method needs to decorate it with a context
+    decorator like :meth:`contextlib.contextmanager`, and needs to yield in
+    the body of the method exactly once. Code that comes before the yield will
+    run before the command starts, and code that comes afterward will run when
+    the command finishes but before the CLI exits.
+
+    For example, if you run MonkeyType against a Django codebase, you can use
+    this method to setup Django before the command runs::
+
+      @contextmanager()
```
carljm: I don't think the parens are needed here?

doc/configuration.rst
```diff
@@ -99,6 +99,33 @@ Subclassing ``Config`` or ``DefaultConfig``
...
 
+  .. method:: cli_context(command: str) -> Iterator[None]
+
+    A context manager which wraps the execution of the CLI command. Use it to
+    perform any initialization or cleanup required by your codebase.
+
+    ``command`` is the command passed to the monkeytype cli: ``'run'``,
+    ``'apply'``, etc.
+
+    Since :meth:`cli_context` is used as a context manager, any :class:`Config`
```
carljm: Using `contextlib.contextmanager` will usually be the simplest way to implement this method, but it's not the only way. You can just as well create a separate context manager class (one with `__enter__` and `__exit__` methods) and return an instance of that class from the method. I appreciate the concern for helping people get this right, but I think right now this documentation is a bit too verbose; goes a little too far in documenting the behavior of `contextlib.contextmanager`, which is out of MonkeyType's scope. I would defer to the Python documentation when we can, rely heavily on the example code (which is the first thing most people will look at anyway) and revise to something shorter:

"This method must return a [context manager](https://docs.python.org/3/reference/datamodel.html#with-statement-context-managers) instance. In most cases the simplest way to do this will be via the [contextlib.contextmanager](https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager) decorator. For example, if you run MonkeyType against a Django codebase, you can setup Django before the command runs::"

doc/configuration.rst
```diff
@@ -99,6 +99,33 @@ Subclassing ``Config`` or ``DefaultConfig``
...
+    For example, if you run MonkeyType against a Django codebase, you can use
+    this method to setup Django before the command runs::
+
+      @contextmanager()
+      def cli_context(self, command: str) -> Iterator[None]:
+          import django
+          django.setup()
+          yield
+
+    The default implementation of this method just calls ``yield`` to let the
```
carljm: "The default implementation of this method just returns a no-op context manager."

doc/configuration.rst
```diff
@@ -99,6 +99,33 @@ Subclassing ``Config`` or ``DefaultConfig``
 
     Defaults to ``False``.
 
+  .. method:: cli_context(command: str) -> Iterator[None]
+
+    A context manager which wraps the execution of the CLI command. Use it to
```
carljm: I'd like to give a bit more context here about _why_ MonkeyType might require this kind of setup/teardown. Maybe like this:

"A context manager which wraps the execution of a CLI command. MonkeyType has to import your code in order to generate stubs for it; in some cases (e.g. Django) some setup is required before code can be imported. Use this method to define the necessary setup or teardown for your codebase."

monkeytype/cli.py
```diff
@@ -235,11 +236,38 @@ def main(argv: List[str], stdout: IO, stderr: IO) -> int:
...
+        elif isinstance(e, RuntimeError):
+            error_msg = f"{config_class}.cli_context() should only yield once, but it yielded multiple times."
+        elif isinstance(e, TypeError):
+            error_msg = f"{config_class}.cli_context() needs to yield once, but it did not yield."
+
+        error_msg = f"{error_msg}\nSee: https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager"
+        raise RuntimeError(error_msg) from e
+
+    if handler_exception:
+        raise handler_exception
```
carljm: This loses the stack trace of the original exception.

monkeytype/cli.py
```diff
@@ -235,11 +236,38 @@ def main(argv: List[str], stdout: IO, stderr: IO) -> int:
...
+
+    handler_exception: Optional[Exception] = None
+    try:
+        with args.config.cli_context(args.command):
+            # Capture the handler exception so it doesn't inadvertently trigger the cli_context() helper errors.
+            try:
+                handler(args, stdout, stderr)
+            except Exception as e:
+                handler_exception = e
+    except (AttributeError, RuntimeError, TypeError) as e:
```
carljm: Again here, definitely appreciate the concern for helping people out when things go wrong, but I think some of the error messages here make assumptions about how `cli_context` must be implemented (with `contextlib.contextmanager`), and about the source of a given exception type, that aren't necessarily valid. `TypeError` or `AttributeError` within their code could have an entirely different source from the one you are anticipating. Fortunately thanks to `raise from` you don't lose the original traceback here, which is good, but my feeling is still that the benefit isn't worth the complexity here, and we're better off relying on Python's own exception messages to be clear enough.

I probably triggered your extra caution here by my previous comment mentioning concern that requiring a context manager from the API might be harder for some developers! I'd rather see if a good example in the docs is enough to avoid that problem, and if not, do more once we have real-world data on what causes confusion.

folz:
Thanks for the guidance. I've simplified the docs and handler.

tests/test_cli.py
```diff
@@ -75,3 +75,11 @@ def test_no_traces(store_data, stdout, stderr):
     assert stderr.getvalue() == "No traces found\n"
     assert stdout.getvalue() == ''
     assert ret == 0
+
+
+def test_cli_context_manager_activated(stdout, stderr):
+    with mock.patch.object(DefaultConfig, 'cli_context') as mock_context:
```
carljm: Instead of patching, let's define a real config with a toy `cli_context` that is an actual context manager that does something (doesn't really matter what, anything you can assert on; append something to a global list, print something to stdout, whatever you like), then assert on the thing the context manager does. This is closer to the real usage pattern and gives better confidence things are working right; patching in tests should be just a fallback for when the alternative is way too painful.

folz:
Makes sense; I changed the test to check against actual behavior and rebased against master. Thanks for the comments!

carljm:
Great work, thanks for the contribution! Merging.

Issue #1, opened 2017-10-27
rewrite _io imports to io


carljm:
e.g. should not generate `from _io import BytesIO` in a stub

iamrajhans:
@carljm can i take this one.

carljm:
@iamrajhans Sure, thank you!

wahajahmed010:
probe

PR #32, opened 2017-12-21
checking if module is _io then changing it to io (#1)

iamrajhans:


iamrajhans:
@carljm ya sure.

carljm:
Perfect, thank you!

PR #33, opened 2017-12-21
TypeVar instances also lack `__qualname__`

vodik:
Sharing a quick fix I needed to get to work on some of the projects I quickly gave this a shot on. It choked on trying to serialize TypeVar instances as it seems like it also lacks `__qualname__`.

### DIFF
--- a/monkeytype/encoding.py
+++ b/monkeytype/encoding.py
@@ -58,6 +58,9 @@ def type_to_dict(typ: type) -> TypeDict:
         qualname = 'Union'
     elif isinstance(typ, _Any):
         qualname = 'Any'
+    # TODO: typeshed stub for TypeVar is object() instead of a type
+    elif isinstance(typ, TypeVar):  # type: ignore
+        qualname = 'TypeVar'
     else:
         qualname = typ.__qualname__
     d: TypeDict = {


vodik:
Not sure how to handle that mypy failure as the expression is, as I understand it, correct. Its just that `TypeVar` is if type `TypingMeta` rather than `object`.

monkeytype/encoding.py
```diff
@@ -2,7 +2,8 @@
 # All rights reserved.
 #
 # This source code is licensed under the BSD-style license found in the
-# LICENSE file in the root directory of this source tree.
+# LICENSE file in the root directory of this source tree. An additional grant
```
carljm: Looks like this came from an older version of this file? Eliminate this addition.

vodik:
> If you can share, I'm curious what kind of code lead to this error. Looks like you're monkeytyping code that passes around TypeVars at runtime? Haven't seen that before.

I dumped out the type variable it was trying to serialize and they're the following: `~T`, `~KT`, and `~VT`, so its tripping over typing internals.

I might be able to share the code, I'd have to ask permission first, but its a pile of asyncio based code built around aiohttp, aiosip, and some internal libraries. The minimum amount of code that I've found that seems to trigger this is handling an aiohttp based request tha'ts formatted by some of our internal code - and that code does define a custom `TypeVar`. Still playing around to see if I can isolate it further.

Another possibility is that the tool is tripping over existing type hints using Mapping/MappingView/KeysView/ItemsView/etc since those are defined in terms of those particular TypeVars, but I can't seem to reproduce this standalone.

> Thanks for the contribution! Can you add a regression test for this?

Sure, I'll get something done up tomorrow morning.

vodik:
Yeah, I think I'm right with the second lead: it's choking with this data being passed into `CallTraceRow.from_trace`:

    CallTrace(<function GenericMeta.__subclasscheck__ at 0x7f8a2e2b3840>, {'self': typing.Type[typing.MutableMapping[~KT, ~VT]], 'cls': typing.Type[generator]}, <class 'bool'>, None)

So I think I know how to generate something that will also duplicate this issue now. But I'll have to dig into that tomorrow. Just sharing in case it makes you think of something.

akx:
I'm also seeing this.

This is the trace being processed:

```python
CallTrace(
  <function ABCMeta.__subclasscheck__ at 0x108a01c80>,
  {
    'cls': typing.Type[typing.MutableMapping[~KT, ~VT]],
    'subclass': typing.Type[pyparsing.ParseResults]}, <class 'bool'>, None),
  }
)
```

carljm:
It looks like in both cases it's choking on a function in `typing.py`. Are you using DefaultConfig with the default code filter? Because if so it should exclude `typing.py` from tracing. If not, maybe switching to that is the quickest workaround. But we can certainly also merge this once it's fixed up.

vodik:
I fixed up the commit in case you decide this needs to be merged now, but I'll get that test added regardless. Need to run some errands to prepare for Christmas,

I suspect it comes from inheriting `MutableMapping` but haven't quite duplicated the error just yet - stopped looking at it yesterday with my last message. I want to get to the root of it so I can write a regression for it.

carljm:
It looks like this fix is wrong in the sense that it doesn't actually allow round-tripping a `TypeVar` instance through serialization. If we have `T = TypeVar('T')` in a module `foo.py`, what we _want_ to serialize is `foo.T` (so that on deserialization we import `T` from `foo` to get it back), but what we will actually serialize is `typing.TypeVar`, so on deserialization we get the `TypeVar` class rather than the TypeVar instance we had originally.

Unfortunately we can't store `foo.T` to enable the normal import-based deserialization, because a `TypeVar` instance doesn't know the module where it was defined. We could special-case TypeVar further such that we also store its name, then we could at least re-create a TypeVar of the same name on rehydration (would require special-casing TypeVar on both the serialization and deserialization ends). This fix would be easily unit-testable, even without being able to create a full tracing repro: just create a TypeVar instance and round-trip it through `type_to_dict` and `type_from_dict` and assert we've still got a TypeVar of the same name.

But it may not be all that important to get this piece right, because generally the user code we want to trace should not be passing around TypeVar instances at runtime. (I doubt that either of you actually cares to generate stub for the `typing` module, so this trace it's choking on isn't even interesting to you.) So the other problem I'm interested in tracking down here is, why does this call trace get recorded in the first place, when it occurs in the standard library, which should be excluded by the default code filter? It's going to be hard for me to track that down without a repro case, but here's some info that could help:

1. Are you (either of you) using a custom MonkeyType config?
2. Assuming no custom config: when you repro in your own codebase, if you can catch the failure in pdb, can you go up the stack trace and figure out how the trace made it through this check? https://github.com/Instagram/MonkeyType/blob/master/monkeytype/config.py#L105

My best guess is that you might have the third-party `typing` module installed as a zipped egg (or maybe you're on a Python install where the entire stdlib is installed zipped??), and so the code object has no `co_filename` and thus the code filter defaults to accepting it.

carljm:
I realized that the underlying issue here is that failure to serialize one weird trace (that we don't even care about!) shouldn't break the entire MonkeyType run. I filed #38 to address that. If we fixed #38, then this issue would become a minor nuisance; just some noise on stderr during the tracing run.

vodik:
Yeah, was just going to comment as as much as TypeVar is probably unserializable, its can't be the only bit that's unserializable.

As for answers to your other questions:

- No custom config
- No, standard lib typing module: `<code object __subclasscheck__ at 0x7fbddf494b70, file "/usr/lib64/python3.6/typing.py", line 1144>`

My hunch currently, and I'll be able to dig into it in depth in a little bit, it something weird is happening when something that subclasses MutableMapping is interacting typing.Dict, but not 100% sure.

carljm:
@vodik To help me track down the default-code-filter problem, can you give me the output of `python -c "from monkeytype.config import LIB_PATHS; print(LIB_PATHS)"` in the same environment you run MonkeyType from? And what OS are you on, so I can try to repro? I think your system Python installation may do something odd with library paths that's breaking the default code filter.

vodik:
Good call, the filter isn't working:

```python
>>> from monkeytype.config import LIB_PATHS; print(LIB_PATHS)
... ('/usr/lib/python3.6', '/home/simon/.virtualenvs/aastrology/lib/python3.6', '<frozen importlib.', '/home/simon/.virtualenvs/aastrology/lib/python3.6/site-packages')
```

Because python is looking in `/lib64`, but that's a symlink to `/lib`. Now the question is if we should improve path matching (I looked at the code, it should be a trivial patch if you're okay moving it to pathlib), or if this is actually another issue in disguise.

> And what OS are you on, so I can try to repro?

Archlinux

vodik:
I drafted a patch that used pathlib to resolve filepaths when filtering, I'll put that up as a separate PR. But I'm still curious what's going on here.

carljm:
Awesome, thanks! Yeah, we should resolve symlinks before path-matching, I don't mind using pathlib for that. If we have that and #39, I think the TypeVar issue becomes much lower priority, but I'm still willing to see if we could do something better here. Still need a repro case to play with in order to understand what that might be.

vodik:
I've almost got something reproducible pulled out - down to three lines, one which is the problem - but its a library call that I can't quite share.

I think its something is going on with my use of a generator to lookup keys into a [bidict](https://github.com/jab/bidict) (a MutableMapping) to then generate a `dict`. I think I should be able to attach a tangible gist soon.

vodik:
@carljm Got it. Its triggered by any call to [`collections.MutableMapping.register`](

```python
import collections


class MyMapping(collections.Mapping):
    _store = {
        'KEY1': 'looked up key 1',
        'KEY2': 'looked up key 2',
        'KEY3': 'looked up key 3',
    }

    def __getitem__(self, key):
        return self._store[key]

    def __len__(self):
        return len(self._store)

    def __iter__(self):
        return iter(self._store)


collections.MutableMapping.register(MyMapping)
```

- [Here][pypasing] is where pyparsing calls it.
- [Here][bidict] is where bidict calls it.

  [pyparsing]: https://github.com/pyparsing/pyparsing/blob/f1d12567a8da4d254e6d62bb0d650c87c7d0bb89/src/pyparsing.py#L943
  [bidict]: https://github.com/jab/bidict/blob/649f1e6f49b996f3dcca929b5f9d3e5bbe0eb203/bidict/_bidict.py#L141

And because of that, I think we're hitting some variation on this issue: https://github.com/python/mypy/issues/1546

vodik:
Closing as this PR is no longer relevant as well. Thanks for the time!

PR #37, opened 2017-12-22
Improve CLI error messages

akx:
This PR makes the CLI error messages a little more explicit.

I was trying to mistakenly pass in file system paths for modules, which didn't work. :)

### DIFF
--- a/monkeytype/cli.py
+++ b/monkeytype/cli.py
@@ -8,6 +8,7 @@ import collections
 import difflib
 import importlib
 import inspect
+import os
 import os.path
 import runpy
 import subprocess
@@ -44,6 +45,11 @@ def module_path(path: str) -> Tuple[str, Optional[str]]:
     parts = path.split(':', 1)
     module = parts.pop(0)
     qualname = parts[0] if parts else None
+    if os.sep in module:  # Smells like a path
+        raise argparse.ArgumentTypeError(
+            f'{module} does not look like a valid Python import path'
+        )
+
     return module, qualname
 
 
--- a/monkeytype/cli.py
+++ b/monkeytype/cli.py
@@ -55,6 +55,14 @@ def module_path_with_qualname(path: str) -> Tuple[str, str]:
     return module, qualname
 
 
+def complain_about_no_traces(args: argparse.Namespace, stderr: IO) -> None:
+    module, qualname = args.module_path
+    if qualname:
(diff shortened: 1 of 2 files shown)


tests/test_cli.py
```diff
@@ -81,7 +81,7 @@ def test_no_traces(store_data, stdout, stderr):
     store, db_file = store_data
     with mock.patch.dict(os.environ, {DefaultConfig.DB_PATH_VAR: db_file.name}):
         ret = cli.main(['stub', func.__module__], stdout, stderr)
-    assert stderr.getvalue() == "No traces found\n"
+    assert stderr.getvalue().startswith("No traces found")
```
carljm: Shouldn't we be able to use `func.__module__` again here and assert on the full error message?

monkeytype/cli.py
```diff
@@ -41,6 +42,11 @@ def module_path(path: str) -> Tuple[str, Optional[str]]:
     parts = path.split(':', 1)
     module = parts.pop(0)
     qualname = parts[0] if parts else None
+    if os.sep in module:  # Smells like a path
```
carljm: Can we add a test for this case?

akx:
@carljm Tests added :)

tests/test_cli.py
```diff
@@ -81,7 +81,7 @@ def test_no_traces(store_data, stdout, stderr):
     store, db_file = store_data
     with mock.patch.dict(os.environ, {DefaultConfig.DB_PATH_VAR: db_file.name}):
         ret = cli.main(['stub', func.__module__], stdout, stderr)
-    assert stderr.getvalue() == "No traces found\n"
+    assert stderr.getvalue() == f"No traces found for module path ('{func.__module__}', None)\n"
```
carljm: I think this `(module, qualname)` representation will be confusing, especially when qualname is `None` (which is the more common case). Can we rejoin the tuple into its original form instead? (Or maybe there is some argparse way to preserve the original pre-parsing string?)

akx:
@mpage Yes, sorry I'd left this hanging! In my defense I spent two weeks in Japan in the interim and [it was pretty](https://www.instagram.com/p/BdfHnvwgKOn/).

@carljm Changes made! :)

mpage:
@akx - Looks great - can you fix the conflicts?

carljm:
Looks great! Thanks for pushing this through! Merging.

PR #39, opened 2017-12-22
Catch and log serialization exceptions

### LINKED ISSUE #38: log and continue when failing to serialize a trace
When collecting traces, we catch unexpected exceptions and just log them and continue execution: https://github.com/Instagram/MonkeyType/blob/master/monkeytype/tracing.py#L254

This is good, because there are lots of strange edge cases where trace collection could fail, and we don't want them to break the MonkeyType run entirely.

But it turns out there are also strange edge cases in trace serialization that can fail (e.g. see #33). It would be good if we were also robust against trace serialization failure; one weird trace failing to serialize shouldn't mean we lose all of our traces.

Given our current logger/store design, it looks like sadly this is something that has to be implemented independently by every trace store, because it's the trace store that is responsible for iterating over a list of traces, serializing them, and storing them.

We could provide a documented `serialize_traces` utility that takes `Iterable[CallTrace]` and returns `Iterable[CallTraceRow]` (or just yields rows, to avoid extra materialization of long lists), and we could implement the exception-catch-log-and-continue there, and then we could use this in the default SQLite store, and it'd be easy for third-party stores to reuse.

carljm:
Fixes #38

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@ Changelog
 master
 ------
 
+* Catch and log serialization exceptions instead of crashing. Fixes #38, merge
+  of #39.
+
 * Fix bug in default code filter when Python lib paths are symlinked. Merge of
   #40. Thanks Simon Gomizelj.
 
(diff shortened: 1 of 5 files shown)


carljm:
Yeah, I thought about that. I agree it's a good idea, but it might take a bit of creativity at this point to do that without breaking backwards compatibility with existing trace stores.

PR #40, opened 2017-12-22
Improve the default code filter's correctness and performance.

vodik:
We need to fully resolve the filename when deciding to filter code because of possible discrepancies that might arise with symlinks (e.g. `/lib64` is symlinked to `/lib` - see #33).

Added a cache around the default filter as well since its essentially pure and adding symlink resolution created an incredibly large performance overhead. Fortunately, with the cache in place, this tool performs even faster than without the patch.

### DIFF
--- a/monkeytype/config.py
+++ b/monkeytype/config.py
@@ -5,6 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 from contextlib import contextmanager
 import os
+import pathlib
 import sys
 import sysconfig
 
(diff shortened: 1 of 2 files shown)


vodik:
I need to step away for a bit, but I'm looking into how I broke `_frozen_importlib` filtering.

carljm:
I would guess it broke because they aren't real paths and pathlib does something weird. Probably should move that handling out of `LIB_PATHS` and special-case it inside the filter instead. Otherwise this looks good!

vodik:
I adjusted it to just drop anything that leads with a '<', so we'll ignore everything that doesn't have a source file associated with it. I don't know how you can meaningfully hint the result of runtime generated code.

carljm:
Looks great! Thanks for paying attention to performance.

vodik:
Looking at this again, as a heads up, I think its probably going to make sense to up the lru_cache maxsize, the 128 default quickly gets exhausted rather quickly.

carljm:
Yeah... hard to pick a value since it depends on the size of the codebase being traced. Maybe I'll try setting it at 8192 for now;. The cache should be pretty memory-efficient, it's just a mapping of a hash to a boolean.

Issue #19, opened 2017-12-15
monketype fails to stub the initial script


Hnasar:
I want to add types to `myscript.py`,

so I do `monkeytype run myscript.py`

but `monkeytype stub myscript.py` says `No traces found`

so I dumped `monkeytype.sqlite3` and see it has traces under the module named `__main__`,

but `monkeytype stub __main__` says `ERROR: Failed decoding trace: Module '__main__' has no attribute 'main'`

carljm:
Yeah, when you run `myscript.py` as a script, its module name is `__main__`; when you run `monkeytype stub myscript`, its module name is `myscript`. It may be possible to hack around this by mangling the name of traces recorded with module `__main__` (if we can figure out the actual module name for the `__main__` module, which might not be easy to do reliably). I have mixed feelings about it; monkeytype needs to import the modules it annotates, and importing a script that isn't carefully written could have unintended side effects. I think I'd rather just have MonkeyType clearly intended for annotating the modules used by the script; if you want to annotate `myscript.py` itself, treat it as a module and make another short script that imports and calls it, and run monkeytype with that script.

For now I'll add a documentation note to that effect. But I'll leave this open as a low-pri enhancement; if someone submits a PR that implements it in a way that looks robust, I'd consider it.

carljm:
Doc clarification added in de7868af5403f4814bcf1c32f61ae697bb2143f1

Thanks for the report! This was definitely an oversight in the documentation, it wasn't at all clear which code `monkeytype run` was intended to trace.

bhrutledge:
> if you want to annotate `myscript.py` itself, treat it as a module and make another short script that imports and calls it, and run monkeytype with that script.

Here's a simple but concrete example of this workaround:

https://github.com/bhrutledge/til/blob/main/python/monkeytype-script.md

ryuta69:
Ohhhhh, it's so sad....

MonkeyType would be great if you can run against single file without workaround.

OTOH workaround is great.

ryuta69:
@bhrutledge maybe this workaround not working anymore..?

I have now this,

```
./
├── src/
│  └── sort.py
└── run_script.py
```

run_script.py
```python
import sys
sys.path.insert(1,'./src')
import sort
sort.main()
```

then, `monkeytype run run_script.py` run correcty.

however, 
```bash
❯ monkeytype -v stub sort
WARNING: Failed decoding trace: No module named 'sort'
[...]

PR #41, opened 2017-12-23
Add support to directly run a module

vodik:
One small feature I would greatly appreciate so I don't have to write my own wrapper scripts: to be able to directly run modules directly just like (`python -m ...`).

I typically heavily lean on entry points scripts and the like when deploying python code, making `runpy.run_script` unpractical.

Feel free to bike shed, or decide you want a different verb, etc, but I'd greatly appreciate it if we can teach MonkeyType how to do this natively.

### DIFF
--- a/monkeytype/cli.py
+++ b/monkeytype/cli.py
@@ -124,10 +124,14 @@ def print_stub_handler(args: argparse.Namespace, stdout: IO, stderr: IO) -> None
 def run_handler(args: argparse.Namespace, stdout: IO, stderr: IO) -> None:
     # remove initial `monkeytype run`
     old_argv = sys.argv.copy()
-    sys.argv = sys.argv[2:]
     try:
         with trace(args.config):
-            runpy.run_path(args.script_path, run_name='__main__')
+            if args.m:
+                sys.argv = sys.argv[3:]
+                runpy.run_module(args.script_path, run_name='__main__', alter_sys=True)
+            else:
+                sys.argv = sys.argv[2:]
+                runpy.run_path(args.script_path, run_name='__main__')
     finally:
         sys.argv = old_argv
 
@@ -195,6 +199,11 @@ def main(argv: List[str], stdout: IO, stderr: IO) -> int:
         'script_path',
         type=str,
         help="""Filesystem path to a Python script file to run under tracing""")
+    run_parser.add_argument(
+        '-m',
+        action='store_true',
+        help="Run a library module as a script"
+    )
     run_parser.add_argument(
         'script_args',
         nargs=argparse.REMAINDER,
--- a/doc/tracing.rst
+++ b/doc/tracing.rst
@@ -33,12 +33,12 @@ script under MonkeyType tracing using ``monkeytype run`` or
 ``monkeytype stub`` and ``monkeytype apply``, to point MonkeyType to the config
 it should use.
 
-Because of the way Python treats scripts and modules differently,
-MonkeyType will record usable traces for modules imported and used by
-the entry point (where ``__name__ == "__main__"``) and not for the
-entry point itself. If you want to annotate the entry point, treat it
-as a module and write another short script that imports and calls its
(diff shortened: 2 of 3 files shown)


vodik:
Amusingly enough, by adding a `__main__.py` and some tweaks to cli argument parsing, monkeytype can become self hosting 😉 . I can add that work too if you guys would like:

```bash
$ monkeytype stub monkeytype.util
from typing import (
    Any,
    Callable,
)


def get_name_in_module(module: str, qualname: str, attr_getter: Optional[Callable[[Any, str], Any]] = None) -> Any: ...
```

monkeytype/cli.py
```diff
@@ -127,7 +127,10 @@ def run_handler(args: argparse.Namespace, stdout: IO, stderr: IO) -> None:
     sys.argv = sys.argv[2:]
     try:
         with trace(args.config):
-            runpy.run_path(args.script_path, run_name='__main__')
+            if args.m:
+                runpy.run_module(args.script_path, run_name='__main__', alter_sys=True)
```
carljm: do we want `run_name='__main__'`? Won't that make traces from the named module unusable?

OTOH, without it we won't trigger `if __name__ == '__main__'` blocks.

vodik: Yes. And then there's actually another problem too. Should there be no import guard inside `__main__.py`, then `monkeytype stub foobar.__main__` launches the application 😦. Maybe this isn't going to be easily workable.

Opinions?

vodik: On the other hand, `python -m` runs it as `__main__` as well anyways. Maybe it makes sense to rewrite away `__main__` when saving the trace?

carljm: I don't have great ideas here. This is why I didn't implement support for this originally; it seemed clearer to maintain a clean separation between the script that exercises the code (which can't be monkeytyped) and the code itself (which can).

I guess the best option, if we're going to do it, is to merge it like this, with the caveat that whatever you run as the script won't be typed. In that case, needs some docs here.

carljm: Hmm, that's a good point. I don't have a problem with rewriting `__main__` to the module name when saving a trace. With #19 I wasn't sure how we'd figure out the right module name, but here the user already gave it to us, so that's not an issue. If you get that working, with docs and tests, I'd happily merge.

vodik: So I thought about this a bit more and there are still some problems but also some potential solutions. We don't really know the module name to rewrite it to, since it could be one of: `foo.py`, `foo.__init__.py`, or `foo.__main__.py`, depending on the module.

We could figure it out with some probing of the filesystem, and possibly some other tricks I haven't thought of, but might not be easily workable.

However another option might be to do it entry points style, e.g. something like the explicit `foo.__main__:main`, and instead of using `run_module` we import `main` from `foo.__main__` and run it ourselves.

carljm: I'm inclined towards what you have now, which is the simplest and least magical implementation, exactly parallel to `python -m`, and just documenting its limitations.

The entry-point style approach would actually be more useful for a package which uses entry points. So you'd give something like `-e foo:main` and we'd internally run the equivalent of:

```
from foo import main
with monkeytype.trace():
    main()
```

This seems like a separate feature from `-m` support. I'm open to having both.

vodik:
Perfect, works for me. I'll get this cleaned up.

vodik:
I added a check before we trace a module to skip anything where `__name__ == "__main__"` as well.

Now to dig into the documentation...

vodik:
And, just for fun, `monkeytype` can be self hosting now 😉:

```
monkeytype run -m monkeytype stub monkeytype.cli
```

carljm:
Looks great! I tweaked the doc wording slightly.

It'd be ideal if we had some end-to-end tests in `test_cli` for `monkeytype run`, but since we have none for the existing functionality, I'm not going to hold up this diff on adding some for `-m`.

Thanks for your work on this!

PR #42, opened 2017-12-23
EHN: modify the example in README

tyrinwu:
I tried to make the example more easy to follow by rearranging the sequence that information is presented. 

First, let users know what the two files contain. Then, start to work on the annotation part.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -34,7 +34,7 @@ Or enable call-trace logging for a block of code in ``myscript.py``::
   import monkeytype
 
   with monkeytype.trace():
-      ...a block of code in myscript.py
+      ...
 
 By default this will dump call traces into a sqlite database in the file
 ``monkeytype.sqlite3`` in the current working directory. You can then use the
--- a/README.rst
+++ b/README.rst
@@ -5,12 +5,10 @@ MonkeyType collects runtime types of function arguments and return values, and
 can automatically generate stub files or even add draft type annotations
 directly to your Python code based on the types collected at runtime.
 
-Examples
---------
+Example
+-------
 
-In the example, we have two files: ``some/module.py`` and ``myscript.py``.
-
-Say ``some/module.py`` originally contained::
+Say ``some/module.py`` originally contains::
 
   def add(a, b):
       return a + b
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -8,17 +8,33 @@ directly to your Python code based on the types collected at runtime.
...
   $ monkeytype run myscript.py
 
-Or enable call-trace logging for a block of code::
+Or enable call-trace logging for a block of code in ``myscript.py``::
 
   import monkeytype
 
   with monkeytype.trace():
-      ...
+      ...a block of code in myscript.py
```
mpage: The `a block of code in myscript.py` comment is redundant. The previous sentence states that this code lives in `myscript.py`.

README.rst
```diff
@@ -8,17 +8,33 @@ directly to your Python code based on the types collected at runtime.
...
+Now we want to infer the type annotation of ``add`` in ``some/module.py`` by
+running ``myscript.py`` with ``MonkeyType``. This is supported in two ways:
+
 Run a script under call-trace logging of functions and methods in all imported
 modules::
 
   $ monkeytype run myscript.py
 
-Or enable call-trace logging for a block of code::
+Or enable call-trace logging for a block of code in ``myscript.py``::
```
carljm: I'm slightly concerned that this gives the wrong impression that the context manager is only for use in a script. The main reason the context manager exists is that in some cases you might want to run tracing on some code that isn't a script at all. I think it might be better to move the context manager down to after the main example, letting the primary example flow more smoothly with only `monkeytype run`.

README.rst
tyrinwu: Thanks for the review. Updated.

README.rst
tyrinwu: You mean that it can also be used in an interactive shell?

 I actually feel this example is a bit too long as well.

tyrinwu:
@carljm @mpage 
I committed a shorter version. I dropped the context manager part because 
   1. I am not sure how to integrate it without cutting the flow of the document.
   2. It seems alright not to talk about it in the beginning.

I moved the limitation part to almost the end since it seems too serious to talk about it at the very beginning.

Sorry for the messy multiple commits.

Issue #9, opened 2017-10-27
Add support for django.utils.functional.cached_property


carljm:

mpage:
Support added in #46.

PR #46, opened 2018-01-04
Add support for django.utils.functional.cached_property

christopherjwang:
Howdy, I think this is what is needed for #9. It passes `pytest`, `flake`, and `mypy`. I'm not sure if I completely mangled the `Pipfile.lock` file though.

screen shot 2018-01-03 at 10 51 05 pm

### DIFF
--- a/monkeytype/util.py
+++ b/monkeytype/util.py
@@ -7,6 +7,11 @@ import importlib
 import inspect
 import types
 
+try:
+    from django.utils.functional import cached_property  # type: ignore
+except ImportError:
+    cached_property = None
+
 from typing import (
     Any,
     Callable,
@@ -46,6 +51,8 @@ def get_func_in_module(module: str, qualname: str) -> Callable:
         else:
             raise InvalidTypeError(
                 f"Property {module}.{qualname} is missing getter")
+    elif cached_property and isinstance(func, cached_property):
+        func = func.func
     elif not isinstance(func, (types.FunctionType, types.BuiltinFunctionType)):
         raise InvalidTypeError(
             f"{module}.{qualname} is of type '{type(func)}', not function.")
--- a/Pipfile
+++ b/Pipfile
@@ -15,6 +15,7 @@ pytest-smartcov = "*"
 cython = "*"
 sphinx = "*"
 twine = "*"
+django = "*"
 
 
 [requires]
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,20 +1,20 @@
 {
     "_meta": {
         "hash": {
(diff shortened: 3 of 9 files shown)


monkeytype/util.py
```diff
@@ -35,17 +40,19 @@ def get_func_in_module(module: str, qualname: str) -> Callable:
     # https://github.com/python/typeshed/blob/master/stdlib/3/inspect.pyi#L213
     func = inspect.unwrap(func)  # type: ignore
     if isinstance(func, types.MethodType):
-        func = func.__func__
+        return func.__func__
```
carljm: Why this change (and below)?

christopherjwang: Personally, I think it's a lot easier to trace. Once we pass certain conditions there's really no point in continuing down into the rest of the function if we can just exit out early with a known result. I think it could also prevent future people from altering `func` later if we explicitly declare the exact object we want to return with an early return statement.

I'm perfectly okay reverting this change if you disagree.

carljm: Let's revert, it's not related to the changes in this diff. It's a judgment call whether later additions to this function are more likely to be things that should or shouldn't also apply to these unwrapped values, I can envision either possibility. If they shouldn't apply, the existing `elif` chain pattern handles that fine. I prefer to err on the side of single return per function, early returns are easy to miss.

carljm:
The reason for the "changes requested" was for tests that exercise the changes to `get_func_in_module` and  `get_func_in_mro`, in addition to the test you added to exercise the change in `stubs.py`.

Thanks again for the pull request!

christopherjwang:
Sure thing! I added the tests and reverted the early returns.

mpage:
@carljm - +1 for pluggable property/descriptor support. That's going to require some refactoring and thought, so I'm on board with merging this as-is.

@christopherjwang - Can you add a test to verify the generated stub for a function that is decorated with `cached_property`? Based on the screenshot, it looks like the `self` argument is annotated with a type, which it shouldn't be.

carljm:
@mpage filed #47 for pluggable property/descriptor support.

Good catch on the self annotation! @christopherjwang can you take a look at why that's happening?

Already-added tests look good.

christopherjwang:
The issue was not including `FunctionKind.DJANGO_CACHED_PROPERTY` in `_KIND_WITH_SELF` [here](https://github.com/Instagram/MonkeyType/pull/46/files#diff-17f06850656e0e2dbfd8fd493eb7402dL74). I've fixed it and added more tests, tests should be everywhere the `@property` decorator is tested.

screen shot 2018-01-06 at 3 53 02 pm

PR #50, opened 2018-01-06
Support sample count

tyrinwu:
This pull request implements the enhancement suggested in issues #7 

It supports `--sample-count` argument for both `stub` and `apply` two subparsers. Currently, it prints "Annotation for module.func based on N call trace(s)." to `stderr` line by line for each method. 

Some points that I am not sure
- `--sample-count` argument is checked at the `get_stub` function. Or it should be checked at `apply_stub_handler` and `print_stub_handler` separately?
-  The two functions, `display_sample_count` and `count_sample` seems to be able to merge into one function. I wrote it separately to test it separately. Should I merge them? 

Some other thoughts 
- add sort by frequency support?

Please let me know if there is anything I miss or any naming issue. I appreciate your review.

### DIFF
--- a/monkeytype/cli.py
+++ b/monkeytype/cli.py
@@ -4,6 +4,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 import argparse
+import collections
 import importlib
 import inspect
 import os.path
@@ -13,6 +14,8 @@ import sys
 import tempfile
 
 from typing import (
+    Callable,
+    Dict,
     IO,
     List,
     Optional,
@@ -27,6 +30,7 @@ from monkeytype.stubs import (
     Stub,
     build_module_stubs_from_traces,
 )
+from monkeytype.tracing import CallTrace
 from monkeytype.typing import NoOpRewriter
 from monkeytype.util import get_name_in_module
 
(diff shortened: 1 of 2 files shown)


mpage:
Thanks for doing this @tyrinwu! 

> --sample-count argument is checked at the get_stub function. Or it should be checked at apply_stub_handler and print_stub_handler separately?

What you have looks good. We can always refactor it later if we add sub-commands that use the `get_stub` function but don't need the added behavior.

> The two functions, display_sample_count and count_sample seems to be able to merge into one function. I wrote it separately to test it separately. Should I merge them?

I would either merge them into a single function or refactor `display_sample_count` so that it takes the output from `count_sample` as an argument.

Can you paste the output when using the `--sample-count` flag so we can get a sense of what this looks like?

tyrinwu:
@mpage  Thanks for looking over!

This is taken from the test case. If there are two functions, `func` and `func2` with 5 traces,

        CallTrace(func, {'a': int, 'b': str}, NoneType),
        CallTrace(func, {'a': str, 'b': str}, NoneType),
        CallTrace(func2, {'a': str, 'b': int}, NoneType),
        CallTrace(func2, {'a': int, 'b': str}, NoneType),
        CallTrace(func2, {'a': str, 'b': int}, NoneType)

It will print to stderr

        Annotation for tests.test_cli.func based on 2 call trace(s).
        Annotation for tests.test_cli.func2 based on 3 call trace(s).

Also, I just merged the two functions into one.

mpage:
Thank you!

tyrinwu:
Thank you. My pleasure!!

PR #52, opened 2018-01-10
Display retype errors

### LINKED ISSUE #49: better debug output when retype application fails
See #48 for a reproducible example case. Running retype directly shows a useful error, but somehow this doesn't get passed through to MonkeyType output, even though we pipe both stdout and stderr in the subprocess call. Need to investigate why not.

mpage:
Catch retype failures and print the error so the user has a better
idea of what went wrong.

Fixes #49.

### DIFF
--- a/monkeytype/cli.py
+++ b/monkeytype/cli.py
@@ -101,6 +101,10 @@ def get_stub(args: argparse.Namespace, stdout: IO, stderr: IO) -> Optional[Stub]
     return stubs.get(module, None)
 
 
+class HandlerError(Exception):
+    pass
+
+
 def apply_stub_handler(args: argparse.Namespace, stdout: IO, stderr: IO) -> None:
     stub = get_stub(args, stdout, stderr)
     if stub is None:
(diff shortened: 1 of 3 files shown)


monkeytype/cli.py
```diff
@@ -274,7 +281,11 @@ def main(argv: List[str], stdout: IO, stderr: IO) -> int:
         return 1
 
     with args.config.cli_context(args.command):
-        handler(args, stdout, stderr)
+        try:
+            handler(args, stdout, stderr)
+        except HandlerError as err:
```
carljm: This is the top level `main`, our final error boundary; is there any reason to be picky about exception types? Should this just be `except Exception as err` so we always return 1 on an exception and print it the same way?

mpage: Generally in agreement with you. My thought was to treat `HandlerError` as an error type that signifies that something went wrong that doesn't represent a bug, whereas other exceptions would represent bugs in our code. I'd like to handle the latter case with a catch-all exception handler that writes the stacktrace to a file in `/tmp` so that users can include it in a bug report. That's obviously a larger change, so I'd prefer to make it in its own PR.

Issue #5, opened 2017-10-27
rewrite NoneType to None


carljm:

m1schka:
not sure if it's related but I also have applied the types and it generated `NoneType` as a parameter, which is unknown to standard python and not getting added to the import (the only option would be to import it from monkeytype but I actually don't want monkeytype imports in my business code).

iamrajhans:
@carljm can i take this one

carljm:
@m1schka Yeah, that's this bug. We have https://github.com/Instagram/MonkeyType/blob/d36c344204eea66aada9f39469f71759d0e68c8f/monkeytype/stubs.py#L327 but it seems it's not effective (at least not in all cases). The correct fix here is that we should never generate `NoneType` at all in an annotation, it should always be translated to `None`.

@iamrajhans Sure, thank you!

johnarnold:
I wrote a test case for this. It happens when types like Optional[str, NoneType] or even more complex occur.
My example:
```python
    def test_union_nonetype_annotation(self):
        """NoneType should always be rendered as None"""
        sig = Signature.from_callable(UpdateSignatureHelper.has_annos)
        sig = update_signature_args(sig, {'a': Dict[str, Dict[str, Union[str, NoneType]]]}, has_self=False,
                                    ignore_existing_annotations=True)
        stub = FunctionStub('test', sig, FunctionKind.MODULE)
        expected = 'def test(a: Dict[str, Dict[str, Union[str, None]]], b) -> int: ...'
        assert stub.render() == expected
```
All the magic happens in render_annotations():
https://github.com/Instagram/MonkeyType/blob/master/monkeytype/stubs.py#L333

I'll submit a PR since this keeps biting me.

Issue #15, opened 2017-12-02
option to generate stubs that ignore existing annotations


carljm:
By default MonkeyType respects existing annotations as the authority, and uses them over annotations inferred from traces when generating stubs. 

For checking correctness of existing annotations against recorded traces, it would be useful to have an option to generate a stub that only considers the traces and ignores existing annotations.

carljm:
Nope, nobody's started work on this that I know of! We were just discussing today how useful it would be, though :) The API should be pretty simple, just a new CLI option like `--ignore-existing-annotations` to `monkeytype stub` command. (It should not be available to `monkeytype apply`; if you ignore existing annotations and try to apply, it'll probably crash with conflicting annotations.)

As far as implementation details, I haven't really looked into it much. If you dive in and have questions, feel free to ask!

tyrinwu:
Thank you so much for the information!! I do feel this would be useful. Will work on it!

carljm:
Fixed with merge of #55

PR #55, opened 2018-01-15
Enhancement: ignore existing annotations

tyrinwu:
This PR implements the enhancement suggested in issue #15 

The `update_signature_args` part is a bit complicated, and I am not sure whether I am doing it right. 
I tested whether a `typ` is `None` or `NoneType`. If so, I assign it as `inspect.Parameter.empty`. I think is related to issue #5 when it regarding parameters. Not sure this is the correct way to do.

I also added one line to  `apply_stub_handler` because I did not add option for `apply_parser` subparser. I need to manually make it `False` and make `get_stub` function aware of the parameter when it is called by `apply_stub_handler`.

```python
 def apply_stub_handler(args: argparse.Namespace, stdout: IO, stderr: IO) -> None:
     args.ignore_existing_annotations = False
```
I appreciate your the review.

### DIFF
--- a/monkeytype/cli.py
+++ b/monkeytype/cli.py
@@ -107,6 +107,7 @@ class HandlerError(Exception):
 
 
 def apply_stub_handler(args: argparse.Namespace, stdout: IO, stderr: IO) -> None:
+    args.ignore_existing_annotations = False
     stub = get_stub(args, stdout, stderr)
     if stub is None:
         print(f'No traces found', file=stderr)
--- a/monkeytype/cli.py
+++ b/monkeytype/cli.py
@@ -95,7 +95,8 @@ def get_stub(args: argparse.Namespace, stdout: IO, stderr: IO) -> Optional[Stub]
     rewriter = args.config.type_rewriter()
     if args.disable_type_rewriting:
         rewriter = NoOpRewriter()
-    stubs = build_module_stubs_from_traces(traces, args.include_unparsable_defaults, rewriter)
+    stubs = build_module_stubs_from_traces(traces, args.include_unparsable_defaults,
+                                           args.ignore_existing_annotations, rewriter)
     if args.sample_count:
         display_sample_count(traces, stderr)
     return stubs.get(module, None)
--- a/monkeytype/cli.py
+++ b/monkeytype/cli.py
@@ -272,6 +272,12 @@ qualname format.""")
         default=False,
         help='Print to stderr the numbers of traces stubs are based on'
         )
+    stub_parser.add_argument(
+        "--ignore-existing-annotations",
+        action='store_true',
+        default=False,
+        help='Ignore existing annotations and generate stubs only from traces.'
+        )
     stub_parser.set_defaults(handler=print_stub_handler)
 
     args = parser.parse_args(argv)
--- a/monkeytype/cli.py
+++ b/monkeytype/cli.py
@@ -99,7 +99,7 @@ def get_stub(args: argparse.Namespace, stdout: IO, stderr: IO) -> Optional[Stub]
         traces,
         include_unparsable_defaults=args.include_unparsable_defaults,
         ignore_existing_annotations=args.ignore_existing_annotations,
-        rewriter=rewriter
+        rewriter=rewriter,
     )
     if args.sample_count:
         display_sample_count(traces, stderr)
@@ -280,7 +280,7 @@ qualname format.""")
         "--ignore-existing-annotations",
         action='store_true',
         default=False,
-        help='Ignore existing annotations and generate stubs only from traces.'
+        help='Ignore existing annotations and generate stubs only from traces.',
         )
     stub_parser.set_defaults(handler=print_stub_handler)
 
--- a/monkeytype/cli.py
+++ b/monkeytype/cli.py
@@ -95,8 +95,12 @@ def get_stub(args: argparse.Namespace, stdout: IO, stderr: IO) -> Optional[Stub]
     rewriter = args.config.type_rewriter()
     if args.disable_type_rewriting:
         rewriter = NoOpRewriter()
-    stubs = build_module_stubs_from_traces(traces, args.include_unparsable_defaults,
-                                           args.ignore_existing_annotations, rewriter)
+    stubs = build_module_stubs_from_traces(
+        traces,
+        include_unparsable_defaults=args.include_unparsable_defaults,
+        ignore_existing_annotations=args.ignore_existing_annotations,
+        rewriter=rewriter
+    )
     if args.sample_count:
         display_sample_count(traces, stderr)
     return stubs.get(module, None)
--- a/monkeytype/stubs.py
+++ b/monkeytype/stubs.py
@@ -181,16 +181,21 @@ def get_imports_for_signature(sig: inspect.Signature) -> ImportMap:
     return imports
 
 
-def update_signature_args(sig: inspect.Signature, arg_types: Dict[str, type], has_self: bool) -> inspect.Signature:
+def update_signature_args(
+        sig: inspect.Signature,
+        arg_types: Dict[str, type],
+        has_self: bool,
+        ignore_existing_annotations: bool = False) -> inspect.Signature:
     """Update argument annotations with the supplied types"""
     params = []
     for arg_idx, name in enumerate(sig.parameters):
         param = sig.parameters[name]
         typ = arg_types.get(name)
-        # Don't touch pre-existing annotations and leave self un-annotated
-        if (typ is not None) and \
-           (param.annotation is inspect.Parameter.empty) and \
-           ((not has_self) or (arg_idx != 0)):
+        typ = inspect.Parameter.empty if (typ is None or typ is NoneType) else typ
+        with_self = (has_self and arg_idx == 0)
+        annotated = param.annotation is not inspect.Parameter.empty
+        # Dont' touch existing annotations unless ignore_existing_annotations
+        if not with_self and (ignore_existing_annotations or not annotated):
             param = param.replace(annotation=typ)
         params.append(param)
     return sig.replace(parameters=params)
@@ -199,11 +204,12 @@ def update_signature_args(sig: inspect.Signature, arg_types: Dict[str, type], ha
 def update_signature_return(
         sig: inspect.Signature,
         return_type: type = None,
-        yield_type: type = None) -> inspect.Signature:
+        yield_type: type = None,
+        ignore_existing_annotations: bool = False) -> inspect.Signature:
     """Update return annotation with the supplied types"""
     anno = sig.return_annotation
-    # Dont' touch pre-existing annotations
-    if anno is not inspect.Signature.empty:
+    # Dont' touch pre-existing annotations unless ignore_existing_annotations
+    if not ignore_existing_annotations and anno is not inspect.Signature.empty:
         return sig
     # NB: We cannot distinguish between functions that explicitly only
     # return None and those that do so implicitly. In the case of generator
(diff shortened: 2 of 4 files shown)


monkeytype/cli.py
```diff
@@ -95,7 +95,8 @@ def get_stub(args: argparse.Namespace, stdout: IO, stderr: IO) -> Optional[Stub]
     rewriter = args.config.type_rewriter()
     if args.disable_type_rewriting:
         rewriter = NoOpRewriter()
-    stubs = build_module_stubs_from_traces(traces, args.include_unparsable_defaults, rewriter)
+    stubs = build_module_stubs_from_traces(traces, args.include_unparsable_defaults,
```
carljm: time to switch to keyword args here to avoid future bugs related to mis-ordered arguments. Also, minor nit but use this indentation style:

```
stubs = build_module_stubs_from_traces(
    traces,
    include_unparsable_defaults=args.include_unparsable_defaults,
    ignore_existing_annotations=args.ignore_existing_annotations,
    rewriter=rewriter,
)
```

monkeytype/stubs.py
```diff
@@ -181,16 +181,21 @@ def get_imports_for_signature(sig: inspect.Signature) -> ImportMap:
...
         param = sig.parameters[name]
         typ = arg_types.get(name)
-        # Don't touch pre-existing annotations and leave self un-annotated
-        if (typ is not None) and \
-           (param.annotation is inspect.Parameter.empty) and \
-           ((not has_self) or (arg_idx != 0)):
+        typ = inspect.Parameter.empty if (typ is None or typ is NoneType) else typ
+        is_self = (has_self and arg_idx == 0)
+        annotated = param.annotation is not inspect.Parameter.empty
+        # Dont' touch existing annotations unless ignore_existing_annotations
```
carljm: `s/Dont'/Don't/`

monkeytype/stubs.py
```diff
@@ -199,11 +204,12 @@ def update_signature_args(sig: inspect.Signature, arg_types: Dict[str, type], ha
...
         sig: inspect.Signature,
         return_type: type = None,
-        yield_type: type = None) -> inspect.Signature:
+        yield_type: type = None,
+        ignore_existing_annotations: bool = False) -> inspect.Signature:
     """Update return annotation with the supplied types"""
     anno = sig.return_annotation
-    # Dont' touch pre-existing annotations
-    if anno is not inspect.Signature.empty:
+    # Dont' touch pre-existing annotations unless ignore_existing_annotations
```
carljm: `s/Dont'/Don't/`

monkeytype/stubs.py
```diff
@@ -239,6 +245,7 @@ def get_updated_definition(
     func: Callable,
     traces: Iterable[CallTrace],
     rewriter: Optional[TypeRewriter] = None,
+    ignore_existing_annotations: bool = False
```
carljm: trailing comma

monkeytype/stubs.py
```diff
@@ -181,16 +181,21 @@ def get_imports_for_signature(sig: inspect.Signature) -> ImportMap:
...
     """Update argument annotations with the supplied types"""
     params = []
     for arg_idx, name in enumerate(sig.parameters):
         param = sig.parameters[name]
         typ = arg_types.get(name)
-        # Don't touch pre-existing annotations and leave self un-annotated
-        if (typ is not None) and \
-           (param.annotation is inspect.Parameter.empty) and \
-           ((not has_self) or (arg_idx != 0)):
+        typ = inspect.Parameter.empty if (typ is None or typ is NoneType) else typ
```
carljm: why did you feel the need to add the check for `NoneType` here? what behavior did you observe without that check that you thought was problematic?

IIUC a `None` here means we recorded no type for this parameter, whereas `NoneType` means we recorded a type of `None`. The former should result in no annotation, the latter should result in an annotation of `None`. So I don't think we should be checking for `NoneType` here.

tests/test_stubs.py
```diff
@@ -426,6 +426,44 @@ def test_update_class(self):
...
+        params = [
+            Parameter('a', Parameter.POSITIONAL_OR_KEYWORD, annotation=str),
+            Parameter('b', Parameter.POSITIONAL_OR_KEYWORD, annotation=bool),
+        ]
+        assert sig == Signature(parameters=params, return_annotation=int)
+
+    def test_update_self_ignore_existing_anno(self):
+        """Don't annotate first arg of instance methods with ignore_existing_annotations"""
+        sig = Signature.from_callable(UpdateSignatureHelper.an_instance_method)
+        sig = update_signature_args(sig, {'self': UpdateSignatureHelper}, True, True)
```
carljm: use keyword args for the last two arguments, too opaque otherwise

tests/test_stubs.py
```diff
@@ -426,6 +426,44 @@ def test_update_class(self):
...
+        expected = Signature(parameters=[Parameter('self', Parameter.POSITIONAL_OR_KEYWORD)])
+        assert sig == expected
+
+    def test_update_arg_ignore_existing_anno_NoneType(self):
+        """Update arg annotations from types"""
+        sig = Signature.from_callable(UpdateSignatureHelper.has_annos)
+        sig = update_signature_args(sig, {'a': NoneType, 'b': int}, False, True)
```
carljm: keyword args for the two boolean arguments

tests/test_stubs.py
```diff
@@ -426,6 +426,44 @@ def test_update_class(self):
...
+        params = [
+            Parameter('a', Parameter.POSITIONAL_OR_KEYWORD, annotation=inspect.Parameter.empty),
+            Parameter('b', Parameter.POSITIONAL_OR_KEYWORD, annotation=int),
+        ]
+        assert sig == Signature(parameters=params, return_annotation=int)
+
+    def test_update_arg_ignore_existing_anno_None(self):
+        """Update arg annotations from types"""
+        sig = Signature.from_callable(UpdateSignatureHelper.has_annos)
+        sig = update_signature_args(sig, {'a': None, 'b': int}, False, True)
```
carljm: again

tests/test_stubs.py
```diff
@@ -426,6 +426,44 @@ def test_update_class(self):
...
+        sig = update_signature_args(sig, {'self': UpdateSignatureHelper}, True, True)
+        expected = Signature(parameters=[Parameter('self', Parameter.POSITIONAL_OR_KEYWORD)])
+        assert sig == expected
+
+    def test_update_arg_ignore_existing_anno_NoneType(self):
+        """Update arg annotations from types"""
+        sig = Signature.from_callable(UpdateSignatureHelper.has_annos)
+        sig = update_signature_args(sig, {'a': NoneType, 'b': int}, False, True)
+        params = [
+            Parameter('a', Parameter.POSITIONAL_OR_KEYWORD, annotation=inspect.Parameter.empty),
```
carljm: I think this test is wrong. `annotation=None` would be correct here.

monkeytype/stubs.py
tyrinwu: Previous, while I tried using `NoneType` in a test. You can test this on the master branch by modifying the `test_update_arg_with_anno` in `test_stubs.py` as follows.

```Python
    def test_update_arg_with_anno(self):
        """Leave existing arg annotations alone"""
        sig = Signature.from_callable(UpdateSignatureHelper.has_annos)
        sig = update_signature_args(sig, {'a': str, 'b': NoneType}, False)
        expected = Signature(
            parameters=[
                Parameter('a', Parameter.POSITIONAL_OR_KEYWORD, annotation=int),
                Parameter('b', Parameter.POSITIONAL_OR_KEYWORD, annotation=NoneType)
            ],
            return_annotation=int
        )
        assert sig == expected
```

I found that with the previous code, when a parameter has a trace suggesting `NoneType`, it will annotate it as `NoneType`. You will pass the test in this way. Thus, I thought that is because we only checked `None` but not `NoneType` in the `update_signature_arg` part, resulting in this behavior. I think as you mentioned in other part, the desire behavior is to return `None` in this way?

monkeytype/stubs.py
tyrinwu: fixed!

monkeytype/stubs.py
tyrinwu: fixed.

monkeytype/stubs.py
tyrinwu: This is the last line, do we need to add a comma?

```
def get_updated_definition(
    func: Callable,
    traces: Iterable[CallTrace],
    rewriter: Optional[TypeRewriter] = None,
    ignore_existing_annotations: bool = False
) -> FunctionDefinition:
```

tests/test_stubs.py
tyrinwu: fixed.

tests/test_stubs.py
tyrinwu: fixed.

tests/test_stubs.py
tyrinwu: fixed.

tests/test_stubs.py
tyrinwu: Related to the discussion in checking `None` and `NoneType` TODO

monkeytype/cli.py
tyrinwu: fixed.

monkeytype/cli.py
```diff
@@ -95,7 +95,12 @@ def get_stub(args: argparse.Namespace, stdout: IO, stderr: IO) -> Optional[Stub]
     rewriter = args.config.type_rewriter()
     if args.disable_type_rewriting:
         rewriter = NoOpRewriter()
-    stubs = build_module_stubs_from_traces(traces, args.include_unparsable_defaults, rewriter)
+    stubs = build_module_stubs_from_traces(
+        traces,
+        include_unparsable_defaults=args.include_unparsable_defaults,
+        ignore_existing_annotations=args.ignore_existing_annotations,
+        rewriter=rewriter
```
carljm: tiny nit: trailing comma

monkeytype/cli.py
```diff
@@ -270,6 +276,12 @@ def main(argv: List[str], stdout: IO, stderr: IO) -> int:
         default=False,
         help='Print to stderr the numbers of traces stubs are based on'
         )
+    stub_parser.add_argument(
+        "--ignore-existing-annotations",
+        action='store_true',
+        default=False,
+        help='Ignore existing annotations and generate stubs only from traces.'
```
carljm: trailing comma

monkeytype/stubs.py
carljm: Yes, the desired behavior when tracing gives `NoneType` is to annotate with `None`, but that's not what your change here does. Your change here causes that parameter to not be annotated at all.

The issue of `NoneType` vs `None` is separate from this enhancement and would best be handled in a separate PR for clarity. This line should only check `if typ is None` and remove the `NoneType` part.

monkeytype/stubs.py
carljm: Yes, our preference is always for trailing comma on every line of a multi-line list. It's not necessary for Python's sake, but it makes future additions or reorderings of the list much easier, with less diff noise.

tests/test_stubs.py
carljm: The fix for this line doesn't seem to be pushed yet?

tests/test_stubs.py
carljm: Yes, `annotation=None` is what we want here, not `annotation=inspect.Parameter.empty`. I think this test can just be removed from this PR because  handling of `NoneType` is not related to handling of `ignore_existing_annotations`. The PR that fixes #5 can add a test similar to this (but doesn't need to use `ignore_existing_annotations`).

tyrinwu:
Thank you for the detailed review. It's good to know the comma preference. I have updated according to your suggestion! (Deleted the extra test and add the trailing comma.)

carljm:
Thanks for your work on this, looks good!

tyrinwu:
Thank you, Carl. I learned a lot from you in this PR.

Issue #58, opened 2018-01-19
Adding command line options like --diff


tyrinwu:
Would you be interested in an option like `git diff` so that users can know the differences between existing annotations and stubs generated only from traces?

This option should only be a sub-option to `stub` I think and can make use of `--ignore-existing-annotations`

Expected output: showing existing annotations and stubs learned only from traces (--ignore-existing-annotations) _alternatively_ **only** for functions that disagree between the twos.

Potential benefits: quickly know where differences are.

carljm:
Yes, I think that'd be a useful and natural thing to build on top of `--ignore-existing-annotations`. I think just `monkeytype stub --diff some.module` would make sense. And I think the most useful output would be literally a diff of the stub generated with ignoring existing annotations and without, e.g. generate the two stubs and use https://docs.python.org/2/library/difflib.html#difflib.Differ to show the difference between them. @mpage does that seem right to you?

tyrinwu:
Will there be cases when one line added some more annotation and thus added one more line and push the output behind it backward by one line? (While in the ignore-existing-annotation counterpart, it does not add one more line.)

If this is the case, it will possibly make using [difflib](https://docs.python.org/2/library/difflib.html#difflib.Differ) produces mismatches between lines?

carljm:
@tyrinwu I'm not totally sure I understand the situation you're suggesting, but a) I don't think we wrap long lines in stub output, so I'm not sure how adding an annotation would bump things down a line, and b) difflib is a smart contextual differ, it's not like the entire diff would be thrown off by a one-line difference near the top.

mpage:
@carljm - What you're suggesting sounds great.

tyrinwu:
@carljm Thanks for these comments. You did understand what I was thinking and it is good to know about difflib is contextual.

mpage:
Fixed by #59

PR #59, opened 2018-01-23
Enhencement: stub --diff

tyrinwu:
This implements issue #58. [`difflib.ndiff`](https://docs.python.org/2/library/difflib.html#difflib.ndiff) is used. In the current implementation, only stubs that differ are shown to users. 

Demo:
```python
def func_anno(a: int, b: str) -> None:
    pass
def func_anno2(a: str, b: str) -> None:
    pass
```
Two traces: 

` CallTrace(func_anno, {'a': int, 'b': int}, int)` <- different from the existing annotation
` CallTrace(func_anno2, {'a': str, 'b': str}, None)`

Output:

```
- def func_anno(a: int, b: str) -> None: ...
?                          ^ -     ^^ ^
+ def func_anno(a: int, b: int) -> int: ...
?                          ^^      ^ ^
```

Demo 2:
```python
def func_anno(a: int, b: str) -> None:
    pass

def super_long_function_with_long_params(
    long_param1: str,
    long_param2: str,
    long_param3: str,
    long_param4: str,
    long_param5: str,
) -> None:
    pass
```
Traces
```
CallTrace(super_long_function_with_long_params, {
    'long_param1': str,
    'long_param2': str,
    'long_param3': int,  <- different
    'long_param4': str,
    'long_param5': int,  <- different
}, None),
CallTrace(func_anno, {'a': int, 'b': int}, int) <-different
```

Output:
```
- def func_anno(a: int, b: str) -> None: ...
?                          ^ -     ^^ ^
+ def func_anno(a: int, b: int) -> int: ...
?                          ^^      ^ ^


  def super_long_function_with_long_params(
      long_param1: str,
      long_param2: str,
-     long_param3: str,
?                  ^ -
+     long_param3: int,
?                  ^^
      long_param4: str,
-     long_param5: str
?                  ^ -
+     long_param5: int
?                  ^^
  ) -> None: ...
```

Please let me know if there is anything need to improve, revise, or delete. I appreciate your review.

### DIFF
--- a/monkeytype/cli.py
+++ b/monkeytype/cli.py
@@ -5,6 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 import argparse
 import collections
+import difflib
 import importlib
 import inspect
 import os.path
@@ -137,12 +138,34 @@ def apply_stub_handler(args: argparse.Namespace, stdout: IO, stderr: IO) -> None
             raise HandlerError(f"Failed applying stub with retype:\n{cpe.stdout.decode('utf-8')}")
 
 
-def print_stub_handler(args: argparse.Namespace, stdout: IO, stderr: IO) -> None:
+def get_diff(args: argparse.Namespace, stdout: IO, stderr: IO) -> Optional[str]:
+    args.ignore_existing_annotations = False
     stub = get_stub(args, stdout, stderr)
-    if stub is None:
-        print(f'No traces found', file=stderr)
-        return
-    print(stub.render(), file=stdout)
+    args.ignore_existing_annotations = True
+    stub_ignore_anno = get_stub(args, stdout, stderr)
+    if stub is None or stub_ignore_anno is None:
+        return None
+    seq1 = stub.render().splitlines(keepends=True)
+    seq2 = stub_ignore_anno.render().splitlines(keepends=True)
+    diff = [s for s in difflib.ndiff(seq1, seq2) if s.startswith(("+", "-", "?"))]
+    for i in range(len(diff) // 4 - 1):
+        diff.insert(5 * (i+1) - 1, "\n")
+    return "".join(diff)[:-1]
+
+
+def print_stub_handler(args: argparse.Namespace, stdout: IO, stderr: IO) -> None:
+    if args.diff:
+        output = get_diff(args, stdout, stderr)
+        if output is None:
+            print(f'No traces found', file=stderr)
+            return
+    else:
+        stub = get_stub(args, stdout, stderr)
+        if stub is None:
+            print(f'No traces found', file=stderr)
+            return
+        output = stub.render()
+    print(output, file=stdout)
 
 
 def run_handler(args: argparse.Namespace, stdout: IO, stderr: IO) -> None:
@@ -286,7 +309,7 @@ qualname format.""")
         "--diff",
         action='store_true',
         default=False,
-        help='Compare stubs generated with or without considering existing annotations.',
+        help='Compare stubs generated with and without considering existing annotations.',
         )
     stub_parser.set_defaults(handler=print_stub_handler)
 
--- a/monkeytype/cli.py
+++ b/monkeytype/cli.py
@@ -149,7 +149,7 @@ def get_diff(args: argparse.Namespace, stdout: IO, stderr: IO) -> Optional[str]:
     seq2 = stub_ignore_anno.render().splitlines(keepends=True)
     diff = [s for s in difflib.ndiff(seq1, seq2) if s.startswith(("+", "-", "?"))]
     for i in range(len(diff) // 4 - 1):
-        diff.insert(5 * (i+1) - 1, "\n")
+        diff.insert(5 * (i+1) - 1, "\n\n\n")
     return "".join(diff)[:-1]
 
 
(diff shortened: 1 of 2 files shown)


monkeytype/cli.py
```diff
@@ -137,12 +138,36 @@ def apply_stub_handler(args: argparse.Namespace, stdout: IO, stderr: IO) -> None
...
-    if stub is None:
-        print(f'No traces found', file=stderr)
-        return
-    print(stub.render(), file=stdout)
+    args.ignore_existing_annotations = True
+    stub_ignore_anno = get_stub(args, stdout, stderr)
+    if stub is None or stub_ignore_anno is None:
+        return None
+    diff = []
+    seq1 = (s + "\n" for s in stub.render().split("\n\n\n"))
```
mpage: Are modules/functions guaranteed to appear in the same order between two calls to get stub? If not, the diff may be incorrect.

mpage:
@tyrinwu - This is great! One inline comment about reordering potentially producing incorrect diffs, but otherwise we're good to go.

monkeytype/cli.py
tyrinwu: This is a very tricky question.  Because in my added function, I split by `\n\n\n`, because this is how they are joint in `ModuleStub`'s `render` function. 

This would involve `render` in both ` ModuleStub` and `ImportBlockStub`. The `render` in `MonduleStub` do sorting on their names. (full code attached below)
```
        if self.imports_stub.imports:  
            parts.append(self.imports_stub.render())
        for func_stub in sorted(self.function_stubs.values(), key=lambda s: s.name):
            parts.append(func_stub.render())
        for class_stub in sorted(self.class_stubs.values(), key=lambda s: s.name):
            parts.append(class_stub.render())
```
In `render` of `imports_stub` (with type: `ImportBlockStub`), it is also sorted. (full code attached below)
```
        for module in sorted(self.imports.keys()):
            names = sorted(self.imports[module])
```
Full code is listed below.

```python
class ModuleStub(Stub):
    def __init__(
            self,
            function_stubs: Iterable[FunctionStub] = None,
            class_stubs: Iterable[ClassStub] = None,
            imports_stub: ImportBlockStub = None
    ) -> None:
        self.function_stubs: Dict[str, FunctionStub] = {}
        if function_stubs is not None:
            self.function_stubs = {stub.name: stub for stub in function_stubs}
        self.class_stubs: Dict[str, ClassStub] = {}
        if class_stubs is not None:
            self.class_stubs = {stub.name: stub for stub in class_stubs}
        self.imports_stub = imports_stub if imports_stub else ImportBlockStub()

    def render(self) -> str:
        parts = []
        if self.imports_stub.imports:
            parts.append(self.imports_stub.render())
        for func_stub in sorted(self.function_stubs.values(), key=lambda s: s.name):
            parts.append(func_stub.render())
        for class_stub in sorted(self.class_stubs.values(), key=lambda s: s.name):
            parts.append(class_stub.render())
        return "\n\n\n".join(parts)

    def __repr__(self) -> str:
        return 'ModuleStub(%s, %s, %s)' % (
            tuple(self.function_stubs.values()), tuple(self.class_stubs.values()), repr(self.imports_stub))

class ImportBlockStub(Stub):
    def __init__(self, imports: ImportMap = None) -> None:
        self.imports = imports if imports else ImportMap()

    def render(self) -> str:
        imports = []
        for module in sorted(self.imports.keys()):
            names = sorted(self.imports[module])
            if module == '_io':
                module = module[1:]
            if len(names) == 1:
                imports.append("from %s import %s" % (module, names[0]))
            else:
                stanza = ["from %s import (" % (module,)]
                stanza.extend(["    %s," % (name,) for name in names])
                stanza.append(")")
                imports.append("\n".join(stanza))
        return "\n".join(imports)

```
Am I on the right track? I believe the order in both sequnces should be the same. The possible mismatch when both sequences are in order is that one sequence is shorter than the other. But I think this should not be possible because they are made from the same traces and the same traces should involve the same functions/classes/import. Please correct me if I oversimplied it.

tyrinwu:
Thanks for the question and the review!! I tried to address the concern.

monkeytype/cli.py
```diff
@@ -137,12 +138,36 @@ def apply_stub_handler(args: argparse.Namespace, stdout: IO, stderr: IO) -> None
...
+    seq2 = (s + "\n" for s in stub_ignore_anno.render().split("\n\n\n"))
+    for stub1, stub2 in zip(seq1, seq2):
+        if stub1 != stub2:
+            stub_diff = "".join(difflib.ndiff(stub1.splitlines(keepends=True), stub2.splitlines(keepends=True)))
+            diff.append(stub_diff[:-1])
+    return "\n\n\n".join(diff)
+
+
+def print_stub_handler(args: argparse.Namespace, stdout: IO, stderr: IO) -> None:
```
mpage: I think we can simplify this to:

```
output, file = None, stdout
if args.diff:
  output = get_diff(args, stdout, stderr)
else:
  stub = get_stub(args, stdout, stderr)
  if stub is not None:
    output = stub.render()
if output is None:
  output, file = 'No traces found', stderr
print(output, file=file)
```

tyrinwu:
@mpage no problem!! Suggestions are good at any time. Thank you for simplifying it!

mpage:
Thanks @tyrinwu!

tyrinwu:
@mpage  thanks for merging it!

Issue #60, opened 2018-01-25
Cli flag to get list of stored modules from sqlite db


msoedov:
Add support for listing the unique set of modules for which we have traces.

Sample usage:

```
> monkeytype list-modules
bar
baz
jaz
```

mpage:
@msoedov - This should be pretty straight forward to implement. Any interest in sending a PR?

msoedov:
@mpage I'd love to!

Issue #26, opened 2017-12-17
Generate all stubs available


thedrow:
I'd like to be able to run `monkeytype stub --all` or something similar to generate all stubs that were recorded.

carljm:
Good idea!

mpage:
@thedrow - How are you imagining that this would be used? For most projects the number of generated stubs will likely be too large to be consumable in a useful manner.

thedrow:
The stubs would be used by the IDE or other tools.

carljm:
@mpage Not all projects are that big :-) I've worked on plenty of projects in the past with module counts in the tens, rather than hundreds or thousands, where it would be reasonable to just want to apply all types and fix all type errors in a single go (but there are still enough modules that it would be a pain to run monkeytype separately for each one.) A couple further thoughts:

1. I can easily envision how this would work for `monkeytype apply`, not so much for `monkeytype stub`, since we currently dump stubs directly to stdout. I think any multi-module output mode would need to create a directory tree on disk instead.
2. I think it might be better to think of it as a recursive mode (like `monkeytype apply somemod --recursive`, which would apply to all submodules as well) rather than `--all`. The latter just makes things really dependent on having the right code filter in place at tracing time, otherwise it gets totally out of hand.

I don't think I'm likely to work on this, but to me it seems interesting enough to warrant leaving the issue open.

johnarnold:
+1 monkeytype apply somemod --recursive

msoedov:
Just a suggestion to use 
```
monkeytype stub .
```
form

carljm:
@msoedov Thanks for the suggestion! Can you clarify what semantics you'd envision for `monkeytype apply .`? Would it try to apply all stubs present in the trace store? Or all stubs for modules within the current working directory (recursively)? The dot suggests the latter to me.

It seems like one downside of this suggestion is that it's less flexible; unlike a `-r` flag, it doesn't generalize to picking any module and applying all stubs within it, recursively. IOW it doesn't have any equivalent for `monkeytype apply some.module -r`.

johnarnold:
@carljm @msoedov 

Is anyone working on --recursive / -r option?  If not, I can take a crack at it. Any opinions on implementation?

carljm:
@johnarnold I'm not aware of anyone working on it, take it away!

hectorv:
How about something like this?
```
❯ monkeytype list-modules | xargs -n1 monkeytype apply
```

[...]

PR #61, opened 2018-01-26
Add list-modules cmd

msoedov:
#60

### DIFF
--- a/monkeytype/cli.py
+++ b/monkeytype/cli.py
@@ -171,6 +171,9 @@ def print_stub_handler(args: argparse.Namespace, stdout: IO, stderr: IO) -> None
 def print_list_handler(args: argparse.Namespace, stdout: IO, stderr: IO) -> None:
     output, file = None, stdout
     modules = args.config.trace_store().list_modules()
+    if not modules:
+        print(f'No traces found', file=stderr)
+        return
     output = '\n'.join(modules)
     print(output, file=file)
 
--- a/monkeytype/cli.py
+++ b/monkeytype/cli.py
@@ -168,7 +168,7 @@ def print_stub_handler(args: argparse.Namespace, stdout: IO, stderr: IO) -> None
     print(output, file=file)
 
 
-def print_list_handler(args: argparse.Namespace, stdout: IO, stderr: IO) -> None:
+def list_modules_handler(args: argparse.Namespace, stdout: IO, stderr: IO) -> None:
     output, file = None, stdout
     modules = args.config.trace_store().list_modules()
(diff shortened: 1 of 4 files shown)


monkeytype/cli.py
```diff
@@ -168,6 +168,16 @@ def print_stub_handler(args: argparse.Namespace, stdout: IO, stderr: IO) -> None
     print(output, file=file)
 
 
+def print_list_handler(args: argparse.Namespace, stdout: IO, stderr: IO) -> None:
```
mpage: Can you rename this to `list_modules_handler` so that it more closely matches the command that it implements?

carljm:
Yeah, I agree with the NotImplementedError plan. We can’t break existing functionality with existing third party backends. It’s ok if the new functionality requires the backend to be extended before it works though. The change log entry should discuss this clearly to avoid confusion from the NotImplementedError.

msoedov:
Good point, I missed to update `CallTraceStore` interface

msoedov:
I have been also thinking about another flag `monkeytype apply_all` or `monkeytype apply *`. However with the current patch it will be possible to achieve the same  results via`monkeytype list-modules | xargs -L monkeytype apply`. WDYT?

johnarnold:
monkeytype apply * is getting really close to a regex implementation... That would be pretty clever and more flexible than just "all", IMO.  Although you could accomplish the same thing with "monkeytype list_all | grep foo_module | xargs monkeytype apply",  I think supporting a regex and a loop internally would probably be faster and certainly more discoverable / easy to document.

carljm:
We've already gotten this feature request once, it's tracked in #26. Personally I still favor a `--recursive` flag; I think this covers the common use cases and is simpler to document and implement. For more advanced use cases you can always pipe (maybe filtered) `list_modules` output to monkeytype (and I don't mind documenting that). I'm not very excited about building module matching by regex into MonkeyType.

johnarnold:
@carljm I'm fine with pipe-based filtering and --recursive too. Just thinking out loud :)

msoedov:
I did not mean using regexp in example  `monkeytype apply *`. `*` is just wildcard value which is kind of apply patch to all modules.  I think even more intuitive would be use  `monkeytype apply .` . Flags as `-r` / `--recursive` is a good option but `monkeytype apply .`  `monkeytype stub .` is easier to remember

carljm:
Looks great! Thanks for doing this, merging. We can continue the other discussion on #26.

msoedov:
My pleasure

Issue #66, opened 2018-01-30
monkeytype apply doesn't apply


johnarnold:
For one of my modules, I ran test cases via monketype run, I can see the traces in the sqlite3 database, and I can run 'monketype stub modulename' and see what looks like valid stub output.

However, when I run 'monkeytype apply modulename', it returns no errors but the module is not modified.

I've been able to successfully update other modules.  This module is my only good repro so far.  There are no errors anywhere. I'm open to ideas on how to proceed. I guess i can start with debugging monkeytype?

johnarnold:
After a little debugging... this seems to happen when there's a mis-match between the directory structure of the module, and the stub when retype is applied.  It doesn't seem to like modules where the the module is a __init__.py in a subfolder of a package.

Manual repro on my module:
```bash
(MonkeyType-sBIHPtkn) johnar@netdev1-westus2:~/github/MonkeyType$ retype --pyi-dir /tmp/monkeytypepn24__qb --target-dir /home/johnar/icm-connector/icm /home/johnar/icm-connector/icm/__init__.py
warning: .pyi file for source /home/johnar/icm-connector/icm/__init__.py not found in /tmp/monkeytypepn24__qb
(MonkeyType-sBIHPtkn) johnar@netdev1-westus2:~/github/MonkeyType$ ls /tmp | grep monkey
monkeytypepn24__qb
(MonkeyType-sBIHPtkn) johnar@netdev1-westus2:~/github/MonkeyType$ ls /tmp/monkeytypepn24__qb/
icm.pyi

**Comment by johnarnold** [comment]
Minimal repro of issue:

files:
https://gist.github.com/johnarnold/438115faedd257519e127d93d1aafc0d
```
repro/
        module1/
                __init__.py
        run.py
```

Repro:
```sh
(MonkeyType-sBIHPtkn) johnar@netdev1-westus2:~/repro$ monkeytype run run.py
(MonkeyType-sBIHPtkn) johnar@netdev1-westus2:~/repro$ monkeytype stub module1
class Foo:
    def __init__(self, arg1: str, arg2: int) -> None: ...
(MonkeyType-sBIHPtkn) johnar@netdev1-westus2:~/repro$ monkeytype apply module1
(MonkeyType-sBIHPtkn) johnar@netdev1-westus2:~/repro$ cat module1/__init__.py
class Foo:
    def __init__(self, arg1, arg2):
        self.arg1 = arg1
        self.arg2 = arg2
```

[...]

PR #69, opened 2018-02-01
Handle __init__.py for modules, print stdout from retype

johnarnold:
Closes https://github.com/Instagram/MonkeyType/issues/66

Generate stubs for modules where the module file is like module/__init__.py

This change preserves the original behavior for modules that are not __init__.py
This change also includes a minor fix to print the stdout/sterr output of retype.
pytests pass.
CLA is signed.


monkeytype/cli.py
```diff
@@ -121,9 +122,16 @@ def apply_stub_handler(args: argparse.Namespace, stdout: IO, stderr: IO) -> None
     mod = importlib.import_module(module)
     src_path = inspect.getfile(mod)
     src_dir = os.path.dirname(src_path)
-    pyi_name = module.split('.')[-1] + '.pyi'
+
     with tempfile.TemporaryDirectory(prefix='monkeytype') as pyi_dir:
-        pyi_path = os.path.join(pyi_dir, pyi_name)
+        if src_path.endswith('__init__.py'):
+            pyi_name = '__init__.pyi'
+            pyi_dir = os.path.join(pyi_dir, module)
```
carljm: This is more complex than it needs to be. We don't need to create a subdirectory at all, we just need to use `__init__.pyi` as the filename. Because when we invoke retype, we pass the final file-containing directory anyway, it doesn't know or care what directories are above that.

monkeytype/cli.py
```diff
@@ -133,7 +141,9 @@ def apply_stub_handler(args: argparse.Namespace, stdout: IO, stderr: IO) -> None
             src_path
         ])
         try:
-            subprocess.run(cmd, shell=True, check=True, stderr=subprocess.STDOUT, stdout=subprocess.PIPE)
+            proc_output = subprocess.run(cmd, shell=True, check=True, stderr=subprocess.STDOUT, stdout=subprocess.PIPE)
+            for line in proc_output.stdout.decode('utf-8').splitlines():
```
carljm: This is great to have for error/warning cases, but does retype have any output in the success case? Can you paste what it is, if so? Just don't want to make things too noisy.

Also, there's no reason to splitlines and then print each line, when we could get the same effect by just printing the whole thing.

Actually, couldn't we just set `stdout=subprocess.STDOUT` above (instead of pipe) and get the same effect?

CHANGES.rst
```diff
@@ -10,6 +10,10 @@ master
 * Add ``--ignore-existing-annotations`` option to ``monkeytype stub``. Merge of
   #55, fixes #15. Thanks Tai-Lin!
 
+* Generate stubs for modules where the module file is like module/__init__.py.
+  Print retype stdout/stderr. Fixes #66.
```
carljm: Also add a reference to the PR number, now that there is one.

monkeytype/cli.py
johnarnold: Yes i'll simplify the printing. I think it evolved that way through iterations and I didn't notice the splitlines+loop.

The success case has no output.

If I set stdout=subprocess.STDOUT, I get an exception:
OSError: [Errno 9] Bad file descriptor

carljm: Oh, maybe `stdout=sys.stdout`? If that doesn't work, then yeah just `print(proc_output.stdout)`

monkeytype/cli.py
carljm: The build is failing because mypy is complaining about `mkdir_p` not being type annotated. But with this simplification we can get rid of `mkdir_p` and make that problem go away :)

johnarnold: Sure.  I thought it might be confusing to not maintain the directory structure, but since it's just a temp folder it should be ok to overwrite files with same names, etc.

carljm: The temp directory doesn't live longer than one invocation of this function (it is cleaned up at the end of the `with` block), and for one invocation of this function we write exactly one file. So there's no possibility of even overwriting anything.

johnarnold: That's how i thought the context manager worked, but still saw folders on disk (?)

monkeytype/cli.py
```diff
@@ -147,9 +151,10 @@ def apply_stub_handler(args: argparse.Namespace, stdout: IO, stderr: IO) -> None
             src_path
         ])
         try:
-            subprocess.run(cmd, shell=True, check=True, stderr=subprocess.STDOUT, stdout=subprocess.PIPE)
+            proc_output = subprocess.run(cmd, shell=True, check=True, stderr=subprocess.STDOUT, stdout=subprocess.PIPE)
+            print(proc_output.stdout.decode('utf-8'), file=stdout)
         except subprocess.CalledProcessError as cpe:
-            raise HandlerError(f"Failed applying stub with retype:\n{cpe.stdout.decode('utf-8')}")
+            raise HandlerError(f"Failed applying stub with retype:\n{cpe.stderr.decode('utf-8')}")
```
carljm: I think this change is wrong, when we pass `stderr=subprocess.STDOUT` to the `subprocess.run` call, we are telling it to redirect stderr to stdout, so it all gets combined on stdout, we will never get anything on stderr. Did you see something specific that made you think this change was needed?

monkeytype/cli.py
```diff
@@ -147,9 +151,10 @@ def apply_stub_handler(args: argparse.Namespace, stdout: IO, stderr: IO) -> None
             src_path
         ])
         try:
-            subprocess.run(cmd, shell=True, check=True, stderr=subprocess.STDOUT, stdout=subprocess.PIPE)
+            proc_output = subprocess.run(cmd, shell=True, check=True, stderr=subprocess.STDOUT, stdout=subprocess.PIPE)
```
carljm: Minor nit, but let's just call this variable `proc`, not `proc_output`; it is a `Process` object and has more than just "output" attributes on it (even though that's all we're using for  now).

tests/test_cli.py
```diff
@@ -51,7 +58,7 @@ def super_long_function_with_long_params(
     pass
 
 
-class LoudContextConfig(DefaultConfig):
+class LoadContextConfig(DefaultConfig):
```
carljm: This was not a typo, it's a "LoudContextConfig" because it prints "loudly" when running. Please revert the change :)

monkeytype/cli.py
carljm: Odd, shouldn't stick around. Maybe if there was a crash inside the context manager?

tests/test_cli.py
johnarnold: Opps! Sure.

monkeytype/cli.py
johnarnold: Sure

tests/test_cli.py
johnarnold: Done.

monkeytype/cli.py
johnarnold: Done.

monkeytype/cli.py
johnarnold: I tested both ways and also tested the subprocess.run redirection several times.  This was the combo that gave me both warning and error outputs, but it's possible I mixed it up.  I'll retest it.

johnarnold: Reverted and tested, still get the warning.

tests/test_cli.py
```diff
@@ -250,7 +257,7 @@ def test_retype_failure(store_data, stdout, stderr):
 
 
 def test_cli_context_manager_activated(capsys, stdout, stderr):
-    ret = cli.main(['-c', f'{__name__}:LoudContextConfig()', 'stub', 'some.module'], stdout, stderr)
+    ret = cli.main(['-c', f'{__name__}:LoadContextConfig()', 'stub', 'some.module'], stdout, stderr)
```
carljm: Need to revert this too, that's what's causing the test failure.

carljm:
Awesome, thanks! Merging.

johnarnold:
Thanks @carljm !

Issue #54, opened 2018-01-11
TypeRewriter design is susceptible to type name collisions


carljm:
Because visitor methods are defined like `rewrite_Dict` and called based on `__name__` of the type, they can be called for unexpected types if there is a name collision. E.g. we have a class named `Dict` that is not `typing.Dict`, and when found in an annotation it results in this crash:

```
  File "/.../lib/python3.6/site-packages/monkeytype/stubs.py", line 574, in build_module_stubs_from_traces 
    defn = get_updated_definition(func, traces, rewriter=rewriter) 
  File "/.../lib/python3.6/site-packages/monkeytype/stubs.py", line 248, in get_updated_definition 
    arg_types = {name: rewriter.rewrite(typ) for name, typ in arg_types.items()} 
  File "/.../lib/python3.6/site-packages/monkeytype/stubs.py", line 248, in <dictcomp> 
    arg_types = {name: rewriter.rewrite(typ) for name, typ in arg_types.items()} 
  File "/.../python3.6/site-packages/monkeytype/typing.py", line 245, in rewrite 
    typ = rw.rewrite(typ) 
  File "/.../lib/python3.6/site-packages/monkeytype/typing.py", line 157, in rewrite 
    return rewriter(typ) 
  File "/.../lib/python3.6/site-packages/monkeytype/typing.py", line 129, in rewrite_Dict 
    return self._rewrite_container(Dict, dct) 
  File "/.../lib/python3.6/site-packages/monkeytype/typing.py", line 123, in _rewrite_container 
    if container.__args__ is None: 
AttributeError: type object 'Dict' has no attribute '__args__' 
```

carljm:
We can fix (or at least work around this) for the container-visiting behavior of the base `TypeRewriter` by adding some isinstance guards, but it remains kind of an inherent design flaw of `TypeRewriter` that could bite others.

carljm:
Hmm, interesting approach in #71. That certainly solves the ambiguity, but at the cost of making it impossible to use the `TypeRewriter` dispatch machinery for any custom type; you'd have to just use `isinstance` checks in `generic_rewrite` instead. That definitely wasn't the design intent. But it might be the best compromise we've got.

The other option would be to totally break back-compat of `TypeRewriter` and fully fix the issue. A couple ways I can think of to do that:

[...]

PR #71, opened 2018-02-12
Make sure we only rewrite typing module types

vodik:
I hit this bug working on a project that uses the ast module. MonkeyType was trying to treat `ast.List` like `typing.List`. And I guess you can't assume *any* of those modules are uniquely named anyways. So add a check to make sure we're looking at the typing module.


vodik:
Deals with #54

carljm:
Thanks for the PR! See my comment on #54 for why I'm not convinced this is the right approach to fix this problem.

Issue #68, opened 2018-01-31
MonkeyType is not working in IPython


tonyfast:
The newer version of MonkeyType does not work in IPython.  I think https://github.com/Instagram/MonkeyType/commit/e8bb8a687e18dbb3c734ef33501a573265465c9d could be the reason.

An interactive session in IPython will `assert __name__ == '__main__'` it would still be meaningful to capture the `CallTrace` so folks could interactively type their notebook/python source.  

I can totally understand if this falls outside the scope of the project.

carljm:
Hi, thanks for the report! Can you give repro steps for how MonkeyType previously worked usefully in an IPython notebook, that fails today? The reason we disabled tracing from `__main__` is that when generating or applying those stubs MonkeyType will try to import `__main__` which doesn't generally work, so I'm not clear how it ever worked in an IPython context. If you can outline a specific usage flow that used to work, we can investigate making it work again. Thanks!

tonyfast:
The use cases are scattered at the moment, but I am hopeful that MonkeyType can be used to provide interactive type annotations why someone is developing interactively.

[The current prototype](https://github.com/deathbeds/XXX/blob/master/XXX/monkey.ipynb) creates cell magics that use the MonkeyType profiler.  The magics will either write the stubs to a file, show the stub in the next input, or show the retyped stub.  I was validating this concept [off of these tests](https://github.com/deathbeds/XXX/blob/master/XXX/tests/test_typing.ipynb).

This code was working with version `17.21.3`.

carljm:
Cool, thanks! Yeah, I can see why this used to work and no longer does. I think it's reasonable for MonkeyType to not record traces from `__main__` by default (because they won't work with the normal stub/apply flow), but that should be configurable, so that someone working with deeper internals (like you are) can still get the traces you need. IOW we should put that in the default code filter instead of hardcoded into `CallTracer`, then you can easily override the default code filter with your own. Sound OK to you?

carljm:
Hmm, it won't be quite that simple. The code filter API only gets the code object, not the frame (which may have been a design mistake); I don't think it's possible to tell from the code object alone whether the module name is `__main__`. So, options that I see:

1. We could just delete that check and go ahead and record traces with module `__main__` by default. I don't think this has any really bad consequences (@vodik do you know of any?); worst is probably a confusing additional entry from the new `list_modules` command.

[...]

PR #72, opened 2018-02-12
The minimum changes to allow __main__ modules to be typed. #68

tonyfast:
This commit moves the filtering logic for __main__ modules to the CallTraceStoreLogger.  @carljm is this what you were thinking?

All of the current tests are passing locally for me and it is back to working order in IPython.  If this is in the right direction I'll start ticking off the contribution guidelines requirements.

### DIFF
--- a/monkeytype/db/base.py
+++ b/monkeytype/db/base.py
@@ -71,7 +71,8 @@ class CallTraceStoreLogger(CallTraceLogger):
         self.traces: List[CallTrace] = []
 
     def log(self, trace: CallTrace) -> None:
-        self.traces.append(trace)
+        if not trace.func.__qualname__.startwith('__main__'):
+            self.traces.append(trace)
 
     def flush(self) -> None:
         self.store.add(self.traces)
--- a/monkeytype/db/base.py
+++ b/monkeytype/db/base.py
@@ -71,7 +71,7 @@ class CallTraceStoreLogger(CallTraceLogger):
         self.traces: List[CallTrace] = []
 
     def log(self, trace: CallTrace) -> None:
-        if not trace.func.__qualname__.startwith('__main__'):
+        if not trace.func.__module__ == '__main__':
             self.traces.append(trace)
 
     def flush(self) -> None:
--- a/tests/db/test_base.py
+++ b/tests/db/test_base.py
@@ -10,7 +10,7 @@ from monkeytype.db.base import CallTraceStoreLogger
 from monkeytype.db.sqlite import (
     create_call_trace_table,
     SQLiteStore,
-    )
+)
 from monkeytype.tracing import trace_calls
 from _pytest.monkeypatch import MonkeyPatch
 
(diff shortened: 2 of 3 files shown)


monkeytype/db/base.py
```diff
@@ -71,7 +71,8 @@ def __init__(self, store: CallTraceStore) -> None:
         self.traces: List[CallTrace] = []
 
     def log(self, trace: CallTrace) -> None:
-        self.traces.append(trace)
+        if not trace.func.__qualname__.startwith('__main__'):
```
carljm: As typecheck on TravisCI is pointing out, this is a typo, should be `__startswith__`.

Also, I think we need to be checking `__module__` not `__qualname__`; `__qualname__` is a bit confusingly named, it's only "qualified" within the module (e.g. includes class name), it doesn't include the module name.

I think these two issues point out the value of including a regression test for this functionality.

carljm:
@tonyfast Looking good! Are you willing to add a quick unittest for the trace store logger verifying this behavior? Thanks!

tonyfast:
@carljm I added the file `testing/db/test_base.py` to test for this condition.  I wasn't really sure the best way to write this test, but I'll wait for your review.

tests/db/test_base.py
```diff
@@ -0,0 +1,53 @@
...
+@pytest.fixture
+def logger() -> CallTraceStoreLogger:
+    conn = sqlite3.connect(':memory:')
+    create_call_trace_table(conn)
+    return CallTraceStoreLogger(SQLiteStore(conn))
+
+
+def test_round_trip(logger):
+    from types import ModuleType
+    module = ModuleType('__main__')
```
carljm: I don't think any of this business with constructing a fake module is really necessary, is it? All that matters is setting the `__module__` attribute on one of your three functions to the string `__main__`.

tests/db/test_base.py
```diff
@@ -0,0 +1,53 @@
...
+    module.func = func3
+    MonkeyPatch().setattr(module.func, '__module__', '__main__', raising=False)
```
carljm: The correct way to use `MonkeyPatch` is not by importing and instantiating it yourself (the clue is that you have to import it from `_pytest`; anything that's supposed to be imported will be in the public `pytest` namespace). It's a fixture, you can request a pre-built instance by just adding `monkeypatch` to the test signature, and then do `monkeypatch.setattr(...)`.

Personally I'd probably just use `from unittest.mock import patch` and then `with patch.object(...)` instead, since that's what we use elsewhere in the tests. But I don't feel strongly, using the `monkeypatch` fixture is OK too.

tonyfast: Thanks for the advice on pytest, I've never used monkeypatch before.

The changes you suggested are updated.

carljm:
Released MonkeyType 18.2.0 with this fix to PyPI.

PR #75, opened 2018-02-17
Never render NoneType in stubs, use None

johnarnold:
fixes https://github.com/Instagram/MonkeyType/issues/5

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@ Changelog
 master
 ------
 
+* Never render NoneType in stubs, substitute None.  Merge of #75, Fixes #5.
+  Thanks John Arnold.
+
 
 18.2.0
 ------
@@ -14,9 +17,6 @@ master
 
 * Generate stubs for modules where the module file is like module/__init__.py.
   Print retype stdout/stderr. Merge of #69, Fixes #66.
-  Thanks John Arnold.
-
-* Never render NoneType in stubs, substitute None.  Merge of #75, Fixes #5.
   Thanks John Arnold.
 
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,9 @@ master
   Print retype stdout/stderr. Merge of #69, Fixes #66.
   Thanks John Arnold.
 
+* Never render NoneType in stubs, substitute None.  fixes #5.
+  Thanks John Arnold.
+
 
 18.1.13
 -------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,7 +16,7 @@ master
   Print retype stdout/stderr. Merge of #69, Fixes #66.
   Thanks John Arnold.
 
-* Never render NoneType in stubs, substitute None.  fixes #5.
+* Never render NoneType in stubs, substitute None.  Merge of #75, Fixes #5.
   Thanks John Arnold.
 
 
--- a/monkeytype/stubs.py
+++ b/monkeytype/stubs.py
@@ -335,20 +335,29 @@ def render_annotation(anno: Any) -> str:
     if _is_optional(anno):
         elem_type = _get_optional_elem(anno)
         rendered = render_annotation(elem_type)
-        return 'Optional[' + rendered + ']'
+        rendered = 'Optional[' + rendered + ']'
     elif hasattr(anno, '__supertype__'):
-        return anno.__name__
+        rendered = anno.__name__
     elif getattr(anno, '__module__', None) == 'typing':
-        return repr(anno).replace('typing.', '')
-    elif anno is NoneType:
-        # This isn't my favorite thing in the world, but by convention NoneType
-        # and None are used interchangeably in signatures.
-        return 'None'
+        rendered = repr(anno).replace('typing.', '')
+    elif isinstance(anno, NoneType):
+        rendered = 'None'
     elif isinstance(anno, type):
         if anno.__module__ in ('builtins',):
-            return anno.__qualname__
-        return anno.__module__ + '.' + anno.__qualname__
-    return repr(anno)
+            rendered = anno.__qualname__
+        else:
+            rendered = anno.__module__ + '.' + anno.__qualname__
+    else:
+        rendered = repr(anno)
+
+    return replace_nonetype(rendered)
+
+
+def replace_nonetype(stub: str) -> str:
+    """Replaces NoneType with None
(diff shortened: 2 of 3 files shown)


johnarnold:
Feedback welcome.  After a night's sleep, I was thinking about inlining the replace_nonetype() function since it's so simple, but the standalone function also sort of serves as documentation/purpose hint.

monkeytype/stubs.py
```diff
@@ -335,20 +335,29 @@ def render_annotation(anno: Any) -> str:
...
+            rendered = anno.__qualname__
+        else:
+            rendered = anno.__module__ + '.' + anno.__qualname__
+    else:
+        rendered = repr(anno)
+
+    return replace_nonetype(rendered)
+
+
+def replace_nonetype(stub: str) -> str:
```
carljm: Yeah, this doesn't deserve to be its own function, just inline it.

tests/test_stubs.py
```diff
@@ -285,6 +285,15 @@ def test_newtype_parameter_annotation(self):
         expected = 'def test(user_id: UserId) -> None: ...'
         assert stub.render() == expected
 
+    def test_union_nonetype_annotation(self):
+        """NoneType should always be rendered as None"""
+        sig = Signature.from_callable(UpdateSignatureHelper.has_annos)
+        sig = update_signature_args(sig, {'a': Dict[str, Dict[str, Union[str, NoneType]]]}, has_self=False,
+                                    ignore_existing_annotations=True)
+        stub = FunctionStub('test', sig, FunctionKind.MODULE)
+        expected = 'def test(a: Dict[str, Dict[str, Union[str, None]]], b) -> int: ...'
```
carljm: This test is good, though really if the full bug were fixed here, the expected annotation would be `Dict[str, Dict[str, Optional[str]]`

tests/test_stubs.py
```diff
@@ -285,6 +285,15 @@ def test_newtype_parameter_annotation(self):
         expected = 'def test(user_id: UserId) -> None: ...'
         assert stub.render() == expected
 
+    def test_union_nonetype_annotation(self):
+        """NoneType should always be rendered as None"""
+        sig = Signature.from_callable(UpdateSignatureHelper.has_annos)
+        sig = update_signature_args(sig, {'a': Dict[str, Dict[str, Union[str, NoneType]]]}, has_self=False,
```
carljm: This is more complex than necessary to reveal the bug; `Dict[str, Union[str, NoneType]]` is sufficient, second level of nesting in a `Dict` isn't necessary.

monkeytype/stubs.py
```diff
@@ -335,20 +335,29 @@ def render_annotation(anno: Any) -> str:
...
-        return anno.__name__
+        rendered = anno.__name__
     elif getattr(anno, '__module__', None) == 'typing':
-        return repr(anno).replace('typing.', '')
-    elif anno is NoneType:
-        # This isn't my favorite thing in the world, but by convention NoneType
-        # and None are used interchangeably in signatures.
-        return 'None'
+        rendered = repr(anno).replace('typing.', '')
+    elif isinstance(anno, NoneType):
```
carljm: The need to both keep this (for top-level `NoneType`) and add the search and replace below, reveals that there is a deeper bug hiding here, see #76. It's not just `None -> NoneType`, but every special case in this function is not applied within generic types, including e.g. `Union[X, NoneType]` -> `Optional[X]`.

monkeytype/stubs.py
```diff
@@ -335,20 +335,29 @@ def render_annotation(anno: Any) -> str:
...
-            return anno.__qualname__
-        return anno.__module__ + '.' + anno.__qualname__
-    return repr(anno)
+            rendered = anno.__qualname__
+        else:
+            rendered = anno.__module__ + '.' + anno.__qualname__
+    else:
+        rendered = repr(anno)
+
+    return replace_nonetype(rendered)
```
carljm: Let's add a comment here: `# Temporary hacky workaround for #76 to fix at least the NoneType issue`, just to clarify for someone in future who is working on the deeper issue.

monkeytype/stubs.py
```diff
@@ -335,20 +335,29 @@ def render_annotation(anno: Any) -> str:
     if _is_optional(anno):
         elem_type = _get_optional_elem(anno)
         rendered = render_annotation(elem_type)
```
carljm: nit: let's not reuse the variable name `rendered` for two different things here, let's just inline this into the next line

CHANGES.rst
```diff
@@ -16,6 +16,9 @@ master
   Print retype stdout/stderr. Merge of #69, Fixes #66.
   Thanks John Arnold.
 
+* Never render NoneType in stubs, substitute None.  Merge of #75, Fixes #5.
```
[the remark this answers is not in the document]
johnarnold: Woops. Fixed.

monkeytype/stubs.py
johnarnold: Fixed.

monkeytype/stubs.py
johnarnold: Agreed. If the nesting problem is handled, the search-replace hack would not be needed.

monkeytype/stubs.py
johnarnold: Done.

monkeytype/stubs.py
johnarnold: done.

tests/test_stubs.py
johnarnold: Agree, but it was my actual repro :)

tests/test_stubs.py
johnarnold: Change it when we fix the other bug?  Or just simplify it now?

carljm: Actually on second thought it would be best if we could adjust this test to use a repro case that doesn't rely on `Union[Something, None]`. Because if we leave it as is, then once #76 is fixed, this test no longer even tests what it's supposed to (avoiding `NoneType`), because there won't be any `None` or `NoneType` here any more, just an `Optional`.

We should be able to repro this with something even simpler like `Dict[str, None]` to keep it totally orthogonal from `Optional` handling.

johnarnold: @carljm simplified it.

carljm:
Awesome, thanks! Merging.

PR #80, opened 2018-03-15
Make cli.py runnable as a stand-alone script

mpage:
Make it so we can run the cli manually via

`python monkeytype/cli.py`


carljm:
I'm curious what workflow led to you wanting to do this, rather than run the installed script?

On a revisit, I'm nervous about committing to support this in the long-term, since it puts all of the internal packages on `sys.path` as top-level, which could easily cause other breakage. In a way the internal `typing` module is a nice fail-fast to make it clear upfront that it's not supported.

Is there another way we could better support bypassing the installed entry point script? Like maybe a top-level script in the repo (`mtcli.py`?) that just imports and calls `entry_point_main`?

mpage:
> Since we expect people to import and subclass TypeRewriter, this is a backwards-incompatible change.  I guess we can go ahead and do it (better now than later) but we'd better at least mention it (probably with an all-caps INCOMPATIBLE CHANGE or similar warning) in the changelog.

Good point! I'd actually prefer to wait for a more substantive change before committing an incompatible change.

> I'm curious what workflow led to you wanting to do this, rather than run the installed script?

No workflow led me to this. I just noticed that it wasn't possible to use the script manually and that seems broken.

> Like maybe a top-level script in the repo (mtcli.py?) that just imports and calls entry_point_main?

👍 

I'm going to abandon this.

carljm:
> I just noticed that it wasn't possible to use the script manually and that seems broken.

It's not broken because `monkeytype/cli.py` is not a script, it's a library module (which happens to contain a function that we use as the entry point for our setuptools-generated script).

Unfortunately Python kind of requires that you maintain a clean separation between scripts and library modules, or else `sys.path` (not to mention `__name__`) gets all kinds of confused.

So yeah, for clarity/ease in development, we should probably add an included script that isn't generated in your virtualenv bin/ dir by setuptools.

Issue #82, opened 2018-04-22
Option to monkeytype installed packages.


BoPeng:
For various reasons, my unittest tests against installed packages and it took me a while to figure out why `monkeytype` does not record any type information. I ended up writing the following `monkeytype_config.py`

```
from types import CodeType
from monkeytype.tracing import CodeFilter
from monkeytype.config import DefaultConfig, default_code_filter

class MyConfig(DefaultConfig):

    def code_filter(self) -> CodeFilter:
        def my_code_filter(code: CodeType) -> bool:
            return 'mypackage' in code.co_filename or default_code_filter(code)
        return my_code_filter

CONFIG = MyConfig()
```
where `mypackage` is a hard-coded name for my pakcage. This works, but it is inconvenient when I try to trace different packages, so I am wondering if MonkeyType can provide a command line option to allow tracing certain packages in `site_packages`.

BoPeng:
I cannot really comment on if this is a "corner case" because all my python packages, especially those involve C/C++ extensions are tested after they are installed to site-packages. I do understand your argument but I would appreciate it if you could document this behavior (ignore site-package) more clearly.

carljm:
Sorry for the delay in getting back to this! I think the behavior of the default code filter is already pretty well documented at https://monkeytype.readthedocs.io/en/stable/configuration.html#monkeytype.config.DefaultConfig.code_filter and https://monkeytype.readthedocs.io/en/stable/tracing.html#deciding-which-calls-to-trace -- I also just added an FAQ entry in 97bb12e6154b2bbe89afcee87cfa3ace995ba690. If there's somewhere else in the documentation you think it ought to be mentioned, let me know!

PR #83, opened 2018-04-25
Allow typing specified modules in site-packages

BoPeng:
#82

This PR allows users to define an environment variable `MONKEYTYPE_MODULE` with name of a module, which would be typed by MonkeyType even if it is under Python site-packages.


BoPeng:
I know the original PR was hacky, it was a proof-of-concept patch just to see if it is potentially acceptable.

I have updated the PR with a better implementation and some doc, but let us see if the following behavior is acceptable before working on a test:

```diff
+    # if MONKEYTYPE_MODULES is defined, type only specified packages
+    if 'MONKEYTYPE_MODULES' in os.environ:
+        # try to remove lib_path to only check package and module names
+        for lib_path in LIB_PATHS:
+            try:
+                filename = filename.relative_to(lib_path)
+                break
+            except ValueError:
+                pass
+        return any(x == filename.stem or x in filename.parts for x in os.environ['MONKEYTYPE_MODULES'].split(','))
+    else:
+        return not any(_startswith(filename, lib_path) for lib_path in LIB_PATHS)
```

Basically,
1. The new variable is called `MONKEYTYPE_MODULES` and allows multiple comma separated module names.
2. In contrast to the original `or` logic where non-site packages are always allowed, **only** specified modules will be typed in this updated PR.
3. The implementation strips `LIB_PATHS` but allows checking any of `package`, `sub-package`, or `module` in filename `lib_path/package/sub-package/module.py`.

carljm:
The proposed implementation sounds great! One bikeshed; let's call the env var `MONKEYTYPE_TRACE_MODULES`, since MonkeyType does multiple things with modules; this env var doesn't affect stub generation or application, only tracing. Similarly, let's make sure we clearly refer to tracing in comments and docs, e.g. "trace only specified packages" not "type only specified packages" in the comment above.

Otherwise, SGTM! Thanks for working on this!

BoPeng:
PR updated. Not sure if the [monkeypatch](https://docs.pytest.org/en/latest/monkeypatch.html)-based test works though.

BoPeng:
Test failed due to the use of lru cache, it is fixed now.

carljm:
Merged, thanks again for the contribution!

BoPeng:
You are welcome, also thanks for open-sourcing this great tool.

BoPeng:
https://github.com/Instagram/MonkeyType/blob/master/doc/faq.rst#i-run-tests-against-my-package-installed-in-site-packages-and-i-get-no-traces also needs to be updated.

carljm:
Oh thanks! Will update now.

carljm:
Btw I just released MonkeyType 18.5.1 with this fix and a couple others merged recently.

BoPeng:
Great. Thanks for letting me know. Updating now.

PR #84, opened 2018-05-11
Add support to rewrite Generator[..., None, None] to Iterator[...] (#4)

iyanuashiri:
Hi @carljm, I have done what you asked me to do. I await a review and feedback. Thank you for the assistance. This PR makes it easy for a user to decide whether to rewrite a Generator[..., None, None] to Iterator[...]

### DIFF
--- a/monkeytype/typing.py
+++ b/monkeytype/typing.py
@@ -251,8 +251,19 @@ class NoOpRewriter(TypeRewriter):
         return typ
 
 
+class RewriteGenerator(TypeRewriter):
+    """Returns an Iterator, if the send_type and return_type of a Generator is None"""
+
+    def visit_Generator(self, typ):
+        args = getattr(typ, '__args__')
+        for e in args:
+            if e.__args__[1] and e.__args__[2] is None:
+                return Iterator[e.__args__[0]]
+
+
 DEFAULT_REWRITER = ChainedRewriter((
     RemoveEmptyContainers(),
     RewriteConfigDict(),
     RewriteLargeUnion(),
+    RewriteGenerator(),
 ))


monkeytype/typing.py
```diff
@@ -251,8 +251,19 @@ def rewrite(self, typ):
         return typ
 
 
+class RewriteGenerator(TypeRewriter):
+    """Returns an Iterator, if the send_type and return_type of a Generator is None"""
+
+    def visit_Generator(self, typ):
+        args = getattr(typ, '__args__')
+        for e in args:
```
carljm: Hmm, this doesn't look quite right to me. I would expect the generator `typ` to have `__args__` like `[Something, None, None]`, which would imply  that we would just want the next two lines, operating on `args` instead of `e.__args__`. Currently we're going two layers deep and I'm not sure why.

monkeytype/typing.py
```diff
@@ -251,8 +251,19 @@ def rewrite(self, typ):
         return typ
 
 
+class RewriteGenerator(TypeRewriter):
+    """Returns an Iterator, if the send_type and return_type of a Generator is None"""
+
+    def visit_Generator(self, typ):
+        args = getattr(typ, '__args__')
```
carljm: this should probably just be `args = typ.__args__`, the getattr doesn't buy us anything

monkeytype/typing.py
```diff
@@ -251,8 +251,19 @@ def rewrite(self, typ):
...
+        for e in args:
+            if e.__args__[1] and e.__args__[2] is None:
```
carljm: this doesn't do what you intend, because `is` binds more tightly than `and`; should be `e.__args__[1] is None and e.__args__[2] is None`

iyanuashiri:
Wow! Thanks. I have implemented the review. I will write the unit tests.

iyanuashiri:
Hey @carljm I am sorry it took this long. Some personal issues affected me. I have written tests but it is failing and I don't know why. Can I send a PR and you do a code review?

carljm:
It seems this is replaced by #110

PR #85, opened 2018-05-13
Fix issue with passing monkeytype cli flags to module

dannyqiu:
When specifying a custom configuration (or using any other command line flag), `sys.argv` is not correct for the script being run:

test.py:
```python
import sys
print(sys.argv)
```

```
> monkeytype -c sample.monkeytype_config:CONFIG run test.py
['test.py', 'run', 'test.py']
```

This fix modifies `sys.argv` based on argument parsing, rather than a hard-coded value.


carljm:
Thanks for the fix!

PR #88, opened 2018-05-23
Print useful error message when filename is passed to stub/apply

### LINKED ISSUE #65: Useful error message if somebody passes a filename instead of module name to stub/apply
We should be able to detect this with a heuristic like "no traces found and os.path.exists(...)"

Shouldn't just use a naive `endswith('.py')`, it's perfectly valid to have a package or module named `py`.

rajathagasthya:
Fixes #65.

### DIFF
--- a/monkeytype/cli.py
+++ b/monkeytype/cli.py
@@ -65,6 +65,11 @@ def complain_about_no_traces(args: argparse.Namespace, stderr: IO) -> None:
     module, qualname = args.module_path
     if qualname:
         print(f'No traces found for specifier {module}:{qualname}', file=stderr)
+    # When there is no trace and a top level module's filename is passed, print
+    # a useful error message.
+    elif os.path.exists(module):
+        print(f"No traces found for {module}; did you pass a filename instead of a module name? "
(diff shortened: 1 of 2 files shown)


monkeytype/cli.py
```diff
@@ -65,6 +65,10 @@ def complain_about_no_traces(args: argparse.Namespace, stderr: IO) -> None:
     module, qualname = args.module_path
     if qualname:
         print(f'No traces found for specifier {module}:{qualname}', file=stderr)
+    # When there is no trace and a top level module's filename is passed, print
+    # a useful error message.
+    elif os.path.exists(module):
+        print(f'{module} does not look like a valid Python import path', file=stderr)
```
carljm: Let's tweak this error message a bit; in fact `foo.py` does "look like a valid Python import path", it just happens to likely not be one in this case. Suggested error message: `f"No traces found for {module}; did you pass a filename instead of a module name? Maybe try just '{os.path.splitext(module)[0]}'."`

carljm:
Oh shoot, my bad! Really wish GitHub notified on new branch pushes and not just on comments. This looks great, thanks @rajathagasthya , sorry for the delay in re-review!

Issue #90, opened 2018-06-28
Running "monkeytype list-modules" without collecting results in Traceback


pradyunsg:
:bug: on v18.5.1

**Demonstration**

```sh-session
$ ls monkeytype.sqlite3
$ monkeytype list-modules
Traceback (most recent call last):
  File "/[path-to-venv]/bin/monkeytype", line 11, in <module>
    sys.exit(entry_point_main())
  File "/[path-to-venv]/lib/python3.6/site-packages/monkeytype/cli.py", line 376, in entry_point_main
    sys.exit(main(sys.argv[1:], sys.stdout, sys.stderr))
  File "/[path-to-venv]/lib/python3.6/site-packages/monkeytype/cli.py", line 361, in main
    handler(args, stdout, stderr)
  File "/[path-to-venv]/lib/python3.6/site-packages/monkeytype/cli.py", line 195, in list_modules_handler
    complain_about_no_traces(args, stderr)
  File "/[path-to-venv]/lib/python3.6/site-packages/monkeytype/cli.py", line 65, in complain_about_no_traces
    module, qualname = args.module_path
AttributeError: 'Namespace' object has no attribute 'module_path'
```

carljm:
Thanks for the report!

pradyunsg:
Is this just a matter of making it print a message and return, instead of calling `complain_about_no_traces`?

carljm:
Yep, looks like it.

PR #106, opened 2018-08-11
Fix error when no module exist (#90)

tyrinwu:
Issue #90 

The error arises because `module_path` used in `complain_about_no_traces` is only added for `stub_parser` and `apply_parser` these two but not the `list_module_parser`.

See: 

https://github.com/Instagram/MonkeyType/blob/629d63d8abe6b0c07750e6dd17a71a809e196370/monkeytype/cli.py#L64-L69

https://github.com/Instagram/MonkeyType/blob/629d63d8abe6b0c07750e6dd17a71a809e196370/monkeytype/cli.py#L293-L302

https://github.com/Instagram/MonkeyType/blob/629d63d8abe6b0c07750e6dd17a71a809e196370/monkeytype/cli.py#L315-L324


carljm:
Merged, thanks for the fix!

PR #107, opened 2018-08-19
Python 3.7 support.

### LINKED ISSUE #78: python 3.7 support
I've updated to python 3.7 beta 2 and get errors with monkeytype

`Python 3.7.0b2 (default, Mar  3 2018, 00:51:31) 
[GCC 6.3.0 20170516] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import monkeytype
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.7/site-packages/monkeytype/__init__.py", line 11, in <module>
    from monkeytype.config import (
  File "/usr/local/lib/python3.7/site-packages/monkeytype/config.py", line 20, in <module>
    from monkeytype.db.base import (
  File "/usr/local/lib/python3.7/site-packages/monkeytype/db/base.py", line 17, in <module>
    from monkeytype.tracing import CallTrace, CallTraceLogger
  File "/usr/local/lib/python3.7/site-packages/monkeytype/tracing.py", line 35, in <module>
    from monkeytype.typing import get_type
  File "/usr/local/lib/python3.7/site-packages/monkeytype/typing.py", line 10, in <module>
    from typing import (
ImportError: cannot import name '_Any' from 'typing' (/usr/local/lib/python3.7/typing.py)
`

carljm:
Fixes #78.

This isn't especially pretty, but it seems to work (i.e. passes all the tests on both Python 3.6 and 3.7). The changes are all due to extensive updates to the internals of the `typing` module in Python 3.7; unfortunately we have to make use of those internals.

Also includes the necessary changes to run tests etc under both Python 3.7 and 3.6 on CircleCI.

### DIFF
--- a/monkeytype/compat.py
+++ b/monkeytype/compat.py
@@ -1,0 +1,45 @@
+from typing import Any
+
+try:
+    # Python 3.7
+    from typing import Union, _GenericAlias  # type: ignore
+
+    def is_any(typ: Any) -> bool:
+        return typ is Any
+
+    def is_union(typ: Any) -> bool:
+        return typ is Union or is_generic(typ) and typ.__origin__ is Union
+
+    def is_generic(typ: Any) -> bool:
+        return typ is Union or isinstance(typ, _GenericAlias)
+
+    def is_generic_of(typ: Any, gen: Any) -> bool:
+        return is_generic(typ) and typ.__origin__ is gen.__origin__
+
+    def qualname_of_generic(typ: Any) -> str:
+        return typ._name or typ.__origin__.__qualname__
+
+    def name_of_generic(typ: Any) -> str:
+        return typ._name or typ.__origin__.__name__
+
+except ImportError:
+    # Python 3.6
+    from typing import _Any, _Union, GenericMeta  # type: ignore
+
+    def is_any(typ: Any) -> bool:
+        return isinstance(typ, _Any)
+
+    def is_union(typ: Any) -> bool:
+        return isinstance(typ, _Union)
+
+    def is_generic(typ: Any) -> bool:
+        return isinstance(typ, (GenericMeta, _Union))
+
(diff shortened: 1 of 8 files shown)


monkeytype/compat.py
```diff
@@ -0,0 +1,45 @@
+from typing import Any
```
mpage: This looks great and I don't think we should change it. But I'm curious why you went with this approach. My first instinct would have been to define a compat "interface" using ABC and then instantiate an instance of that based on which minor version of Python was running. I guess with this approach we always ensure that the implementation details that we rely on are present rather than assuming that they are.

carljm: I didn't design it this way so much as just evolve it; initially I thought all I needed was a few import shims, then as it became clear I needed alternate implementations of some queries, I just stuck them into the same format. I also thought this could maybe use some cleanup, just ran out of time; that's partly why the PR title says WIP still.

In terms of what we should do, there are a couple orthogonal issues:

1) Define a proper interface and then inherit two versions of it. This seems like probably a clearer way to structure it, regardless of how we decide which version to use.

2) Version-checking vs catching ImportError. In the past I've leaned towards the latter, because it seems like a more direct check of the thing you really care about (and in some cases plays better with someone using an intermediate/development version for whatever reason). But in this case the distinguishing import is pretty subtle (`_GenericAlias`), and I agree it would probably be clearer to version-check.

I'll clean this up.

carljm: Changed my mind after looking at it a bit, I'm going to leave it as-is. These are all standalone functions, I don't think it's worth the boilerplate to make them all classmethods/staticmethods of some class that's just a namespace. I could make the switch to `sys.version` checking but I'm not sure that matters enough, just going to merge this as-is for now.

PR #108, opened 2018-08-31
Fix a ResourceWarning: unclosed file in test_cli.py

BoboTiG:
Hello,

A little fix to ensure there is no resource leak.


carljm:
Thanks! Updated to just use context manager, if that passes tests will merge.

Issue #102, opened 2018-07-25
demo example tests are not passing at 100%


bluebird75:
I tried to run the demo but I got:

```
monkeytype\demo>pytest
============================= test session starts =============================
platform win32 -- Python 3.6.1, pytest-3.5.0, py-1.5.3, pluggy-0.6.0
rootdir: monkeytype\demo, inifile: pytest.ini
plugins: cov-2.5.1
collected 8 items

test_inbox.py ......F.                                                   [100%]

================================== FAILURES ===================================
_______________________________ test_everything _______________________________

    def test_everything():
        u = make_user()
        other = make_user(name="Other", following=[u.id])
        first_entry = make_feedentry(user_id=u.id, caption="My First Post")
        follow = make_followed(user_id=u.id, follower_id=other.id)
        second_entry = make_feedentry(user_id=u.id, caption="Second Post")
        like1 = make_liked(user_id=u.id, liker_id=other.id, feedentry_id=first_entry.id)
        comment = make_commented(
            user_id=u.id, commenter_id=other.id, feedentry_id=first_entry.id
        )
        like2 = make_liked(user_id=u.id, liker_id=other.id, feedentry_id=second_entry.id)
        repo = FakeRepo(u, other, first_entry, second_entry, like1, like2, comment, follow)
        box = inbox.Inbox(u, repo)

>       assert (
            box.aggregate()
            == [
                models.AggregatedItem(
                    type=models.EventType.LIKED,
                    text='Other liked your post "Second Post".',
                    published=like2.published,
                ),
                models.AggregatedItem(
                    type=models.EventType.COMMENTED,
                    text="Other commented on your post.",
                    published=comment.published,
                ),
                models.AggregatedItem(
                    type=models.EventType.LIKED,
                    text='Other liked your post "My First Post".',
                    published=like1.published,
                ),
                models.AggregatedItem(
                    type=models.EventType.FOLLOWED,
                    text="Other started following you.",
                    published=follow.published,
                ),
            ]
        )
E       assert [AggregatedIt... 26, 231564))] == [AggregatedIte... 26, 231564))]
[...]

PR #109, opened 2018-09-02
Add workaround to Windows poor datetime accuracy to make tests succeed

bluebird75:
Workaround for problem detected in https://github.com/Instagram/MonkeyType/issues/102

This will be necessary anyway to have a full appveyor build because demo tests are run as part of the regular tests (see the output of pytest -v)

### DIFF
--- a/demo/test_inbox.py
+++ b/demo/test_inbox.py
@@ -4,6 +4,7 @@
 Sample code for MonkeyType demonstration exercise at PyCon 2018.
 
 """
+import sys
 from datetime import datetime, timedelta
 from typing import Collection, Dict, List, Optional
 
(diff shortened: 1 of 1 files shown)


demo/test_inbox.py
```diff
@@ -64,13 +64,19 @@ def make_user(**kwargs):
     return models.User(**defaults)
 
 
+def now():
+    # Workaround for Windows where two close call to datetime.now() return
+    # exaclty the same datetime
+    return datetime.now() + timedelta(microseconds=last_auto_id)
```
mpage: Please guard this with a conditional so that the workaround is only applied when executing on a windows platform.

demo/test_inbox.py
```diff
@@ -64,13 +64,19 @@ def make_user(**kwargs):
     return models.User(**defaults)
 
 
+def now():
+    # Workaround for Windows where two close call to datetime.now() return
+    # exaclty the same datetime
```
mpage: Nit: s/exaclty/exactly/

bluebird75:
Should be good now.

PR #110, opened 2018-09-06
Add support to rewrite Generator[..., None, None] to Iterator[...] (#84)

iyanuashiri:


monkeytype/typing.py
```diff
@@ -251,8 +251,18 @@ def rewrite(self, typ):
...
 
 
+class RewriteGenerator(TypeRewriter):
+    """Returns an Iterator, if the send_type and return_type of a Generator is None"""
+
+    def visit_Generator(self, typ):
+        args = typ.__args__
+        if args[1] is None and args[2] is None:
+            return Iterator[args[0]]
+
```
mpage: This should return `typ` in the case where no rewrite occurs.

mpage:
Also, please fix the broken test. Thanks!

carljm:
I'm confused which PR is current between this one and #84. @iyanuashiri can you close whichever one you don't intend to pursue?

monkeytype/typing.py
iyanuashiri: I have fixed the broken tests. Thanks

iyanuashiri:
@carljm I just fixed this PR. Thanks to you all @mpage

monkeytype/typing.py
```diff
@@ -251,8 +251,19 @@ def rewrite(self, typ):
         return typ
 
 
+class RewriteGenerator(TypeRewriter):
+    """Returns an Iterator, if the send_type and return_type of a Generator is None"""
+
+    def rewrite_Generator(self, typ):
+        args = typ.__args__
+        if issubclass(args[1], NoneType) is True and issubclass(args[2], NoneType) is True:
```
carljm: this should just be `if args[1] is None and args[2] is None`.

`is True` is redundant and should be omitted in both cases, and `is None` is better than `issubclass(..., NoneType)`

carljm: Or wait, I see that's what you had before; are these actually `NoneType` and not `None`? if so, you should still omit `is True`, and use `is NoneType` instead of `issubclass(..., NoneType)`

iyanuashiri:
I am so glad for the merge. Can I add my name to the list of contributors? Can you explain why `if args[1] is NoneType` is better than `issubclass(args[1], NoneType)`?

carljm:
We don't have any list of contributors other than the git history, so you are already there! I also added your change to the changelog in `CHANGES.rst`, with thanks to you for the PR.

> Can you explain why if args[1] is NoneType is better than issubclass(args[1], NoneType)

Because it's simpler and shorter, and there is no chance of encountering a subclass of `NoneType` here. In general in Python it's best to use `is` for comparisons with singletons like `None` (and `NoneType` is also a singleton).

iyanuashiri:
Thanks

Issue #105, opened 2018-07-31
record per-key types for heterogenous dicts and suggest TypedDict annotation


carljm:
Currently MonkeyType assumes dicts are homogenous. When heterogenous dicts are seen, we just combine all the value types into a Union (and then if that Union gets too big, by default we rewrite it to Any). It'd be more useful in such cases to suggest a TypedDict annotation with per-key value types.

The trick here is that we don't really know the intention, so there'd be some level of heuristic involved. Like, if there's multiple non-None value types (and probably if the number of keys is less than some threshold), rather than recording just a single value type, we record per-key value types. Then we'd also need some smarts when combining traces into an annotation: if we have any traces with per-key value types, try to construct a consistent TypedDict, but if we find that key names and value types aren't matching up consistently, fallback to normal Dict.

There's enough fuzziness in the heuristics here and this'd be enough work that it might not be worth it, but for codebases that use lots of heterogenous dicts, it'd make the generated annotations a lot more useful.

devdoomari3:
how about having a wrapper that 'acts like a dict' and shows intention-to-be-TypedDict ?

eg). TypedDictPlaceHolder: https://gist.github.com/devdoomari3/3e6e9e8177d2ff97c4ef0091209984f5

devdoomari3:
@carljm my co-worker @Anylee2142 came up with an improvisation:

- monkeytype extracts 'type-details' for dictionaries
- custom clients can do whatever they want from 'type-details'

```
{
'qualname' : 'Dict', 'elem_types': [... some existing things ...],
'type_details': {
     name: str,
     age: int,
     id: int,
     ...
     }
}
```

what do you think about this?

carljm:
@devdoomari3 Yes, recording `type_details` is what's needed; just need to be careful not to drown the backing store if we run into some massive dicts, so there should probably be a dict size threshold beyond which we don't record details.

And we would want MonkeyType itself to automatically generate TypedDict annotation if it finds a dict for which the set of keys and their types is "consistent enough" across traces, not just leave that to custom clients.

One interesting choice is whether to ever generate `total=False` TypedDict, or always bail back to regular Dict if the set of keys isn't consistent across all traces. For simplicity in the initial version I'd probably lean towards only generating TypedDict if the same set of keys are always present.

Thanks for working on this issue!

carljm:
> I personally don't believe in heuristics though

[...]

PR #124, opened 2018-12-04
WIP: Add StructuredDict for heterogeneous dicts

devdoomari3:
related: https://github.com/Instagram/MonkeyType/issues/105

this pull-request adds `StructuredDict` to allow users to signal their intention to have a dict 'structurally typed'

ps: this is my first commit on MonkeyType, so I'd really appreciate some feedbacks (hence WIP)


carljm:
Hi @devdoomari3 ! Thanks very much for the idea and the pull request, but I don't think we want to take this approach to the problem. Switching your code to use a subclass of `dict` at runtime can be risky (exact type equality checks that passed before could fail, e.g. if using `msgpack` with strict types, performance can suffer), and I don't think there's a need for it here. The goal of MonkeyType is to generate best-effort annotations _without_ requiring changes to your code. By the time you've gone and figured out which dicts in your code should be StructuredDict, you may as well just go ahead and create the TypedDict type and annotate using it. The other problem is that if you use StructuredDict, then let MonkeyType annotate as a TypedDict, you'll immediately get typecheck errors after annotating because the typechecker won't consider StructuredDict a compatible type with the TypedDict annotation.

I think we can get pretty useful behavior with heuristics as described in #105 (can make the size threshold for storing per-key types in tracing customizable, and people can use TypeRewriters to customize the output heuristics to their satisfaction), and we should go that route.

If you're interested in updating this PR to take that approach, that would be great! Let me know and I'll reopen.

devdoomari3:
@carljm I was thinking of re-placing `StructuredDict` to `TypedDict` after running Monkeytype.
(so after running MonkeyType, we won't see `StructuredDict` on the source file)

PR #126, opened 2018-12-20
add type-metadata on CallTrace

devdoomari3:
rel: https://github.com/Instagram/MonkeyType/issues/105

adds type-metadata

ps: will rebase


devdoomari3:
`pipenv install dataclasses` freezes...

anyway, I'll try to do it using NamedTuple

carljm:
Hi @devdoomari3 ! Thanks for working on this issue.

I'm having a bit of trouble following some of the design decisions made in this PR, and it feels like unrelated stylistic changes (e.g. switching to keyword arguments in some CallTrace instantiations) and code reorganizations (moving existing code to different modules) have been included, which makes it harder to review. In the future when submitting your first PR to a new project, I would suggest making a strong effort to minimize the size of the diff you are submitting to only the absolutely necessary changes; this will make it a lot easier for the project maintainer(s) to understand and review your diff. If you feel that stylistic or organizational changes are warranted, submit those in separate pull requests with their own independent justification.

One key design decision issue: I'm not sure why we need to add new columns to the SQLite table (which introduces compatibility problems with people upgrading MonkeyType when they have existing type databases), when the existing `arg_types` and `return_type` and `yield_type` columns already store arbitrary JSON structures and we can just extend the format of that JSON in the existing columns instead.

I actually need this feature to exist in the near future, so I'm going to work on implementing it in a separate PR. If I'm able to reuse some of your work from this PR I will definitely make sure to credit you for it in the commit message and changelog! Thanks again for your work, and sorry for the slow review; I was AFK for much of the last month.

PR #128, opened 2019-01-17
Generate ... for all defaults, rip out defaults-related options.

### LINKED ISSUE #123: Generate '...' instead of default value
From what I can see in [typeshed](https://github.com/python/typeshed/blob/master/stdlib/3/builtins.pyi#L310) and the [mypy docs](https://mypy.readthedocs.io/en/latest/stubs.html#stub-file-syntax), default values for callables' parameters should be replaced with `...`.

This would also obviate the need for '--(in|ex)clude-unparsable-defaults', assuming you could then replace the `object` type with `Any` in the stub output.

carljm:
Fixes #123

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,11 @@ Changelog
 master
 ------
 
+* Render `...` for all parameter defaults in stubs. Remove the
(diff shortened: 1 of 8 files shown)


carljm:
Merged as part of #129.

PR #146, opened 2019-11-05
Fix crash on name collisions with typing containers

ambv:
Due to dynamic method dispatch in the TypeRewriter, any type that matched the names List, Set, Tuple, etc. would crash for missing `__args__`.

We fix this by checking if the module matches, and then using `getattr()` anyway, in case there was a custom `typing` module used.

test_typing.py is a bit ugly now due to the name conflict.

[depends on #145, ignore the first commit please]


carljm:
Merged in fb5b16d31cafb46effc1329aa5dd4b362a9fc204.

PR #149, opened 2019-11-05
Allow for importing modules instead of bare types

ambv:
Some projects, EdgeDB in particular, avoid importing bare names to avoid
conflicts.  Instead, they rely on importing modules and addressing type names
with a proper module or subpackage prefix.

This change provides a `--import-modules` CLI option that changes the default
behavior from:
```py3
  from package.subpackage import SomeType

  def func(arg: SomeType) -> None: ...
```
into:
```py3
  from package import subpackage

  def func(arg: subpackage.SomeType) -> None: ...
```
This change leaves imports from top-level modules alone, so for example `from
typing import List` stays as it was.  This can be further refined if needed.

[Depends on #145, #146, #148; ignore the first three commits please]

### DIFF
--- a/monkeytype/cli.py
+++ b/monkeytype/cli.py
@@ -28,6 +28,7 @@ from monkeytype.config import Config
 from monkeytype.exceptions import MonkeyTypeError
 from monkeytype.stubs import (
     ExistingAnnotationStrategy,
+    ImportStrategy,
     Stub,
     build_module_stubs_from_traces,
 )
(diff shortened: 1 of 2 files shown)


ambv:
Re 1: We're actually using `from typing import *` so it was just handy not to meddle with top-level imports, esp. that there's few types imported that way and they are core enough to be unambiguous. I guess we can also add a pure version of --import-modules that also imports top-level modules. Then we'd have three: import types, import types for top-level modules only, import modules only.

Re 2: Right.

carljm:
I guess I wonder if "import types for top-level modules only" is generally useful or idiosyncratic to EdgeDB. Which makes me think perhaps we should just delegate to a config method for this so that you can implement whatever strategy you like (while the default would remain import-types, and perhaps we'd document what it looks like to implement always-import-modules). WDYT?

It does seem to me on second look that either way (whether the config can be a fully custom implementation, or just returns a strategy enum) this should be on the config rather than a CLI option. Unlike the existing-annotations strategy, this is not something you're likely to want to flip around per-invocation of `monkeytype stub`. Rather, it's a property of your codebase that should be defined once in your config.

carljm:
Closing due to inactivity.

PR #153, opened 2019-11-19
add missing depedencies to setup.py

bollwyvl:
Congratulations on the release, and thanks for MonkeyType!

Looks like there are more typeddicts (yay!) and more consistent casing (yay!), but those libraries weren't added to the dependencies (boo!).

Adding a thorough test, i suppose, would require doing a non-pipenv install of the library during CI... but ~~happy to add if so desired~~.

c/f https://github.com/conda-forge/monkeytype-feedstock/pull/11

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ setup(
         ]
     },
     python_requires='>=3.6',
-    install_requires=['retype'],
+    install_requires=['mypy_extensions', 'retype', 'stringcase'],
     classifiers=[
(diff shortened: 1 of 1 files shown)


bollwyvl:
Thanks, @facebook-github-bot, but I won't be signing the CLA for a one-line fix. I do hope it gets fixed, however!

carljm:
Oof. Thanks for the report! Fixed in 19f7449094210ad5a75846dd38d554acf47cc8c1 and I'll release 19.11.1.

PR #162, opened 2020-02-03
Handle TypedDicts nested within generic types.

### LINKED ISSUE #159: Invalid import when list contains dict
Given the following code:
```python
def foo(a):
    print(a)

foo([{"a": "b"}])
```
When it gets annotated by monkeytype the following will be the result:
```python
from monkeytype.encoding import DUMMY_NAME
from typing import List
def foo(a: List[DUMMY_NAME]) -> None:
    print(a)

foo([{"a": "b"}])
```

This raises `ImportError` as there's no `DUMMY_NAME` in `monkeytype.encoding`.

I'm using the latest monkeytype which is `MonkeyType==19.11.2`.

pradeep90:
Fixes #159.

Earlier, we didn't pass in `max_typed_dict_size` recursively for all calls to `get_type`. This meant that `[{"a": 1}]` would produce the TypedDict class name even when `max_typed_dict_size` was set to 0. And since we didn't gather class stubs from nested generic types, the class definition stub wouldn't show up either.

* Pass `max_typed_dict_size` in all calls of `get_type`.

* Gather class stubs recursively using a `TypeRewriter` visitor for parameter types, return types, and yield types. Add `rewrite_TypedDict` method to `TypeRewriter`.

* Special-case `{}` so that we don't return an empty TypedDict regardless of `max_typed_dict_size`.


monkeytype/stubs.py
```diff
@@ -503,22 +503,62 @@ def __repr__(self) -> str:
...
+class ReplaceTypedDictsWithStubs(TypeRewriter):
+    """Replace TypedDicts in a generic type with class stubs and store all the stubs."""
+
+    def __init__(self, class_name_hint: str) -> None:
+        self._class_name_hint = class_name_hint
+        self._stubs: List[ClassStub] = []
+
+    @property
+    def stubs(self) -> List[ClassStub]:
+        return self._stubs
```
carljm: What's the purpose of this indirection?

monkeytype/stubs.py
```diff
@@ -503,22 +503,62 @@ def __repr__(self) -> str:
...
+
+    def _rewrite_container(self, cls: type, container: type) -> type:
```
carljm: Typically it's preferable to use `typing.Type` (with a subscript clarifying which types are accepted) rather than `type` for type annotations. But it looks like our `typing.pyi` stub forces your hand here, so that's probably an orthogonal change.

monkeytype/stubs.py
```diff
@@ -503,22 +503,62 @@ def __repr__(self) -> str:
...
+            return container
+        args = getattr(container, '__args__', None)
+        if args is None:
+            return container
+        elif args == ((),):  # special case of empty tuple `Tuple[()]`
+            elems = ()
+        else:
+            # Avoid adding a suffix for the first one so that
+            # single-element containers don't have a numeric suffix.
+            elems, stub_lists = zip(*[  # type: ignore
```
carljm: What is the type error being ignored here?

monkeytype/stubs.py
```diff
@@ -503,22 +503,62 @@ def __repr__(self) -> str:
...
+                self.rewrite_and_get_stubs(
+                    elem,
+                    class_name_hint=self._class_name_hint + (
+                        '' if index == 0 else str(index + 1)))
+                for index, elem in enumerate(args)])
+            for stubs in stub_lists:
+                self._stubs.extend(stubs)
+        return cls[elems]  # type: ignore
```
carljm: What is the type error being ignored here?

monkeytype/typing.py
```diff
@@ -116,19 +116,24 @@ def make_generator(yield_typ, send_typ, return_typ):
...
 
-def get_dict_type(dct, max_typed_dict_size=None):
+def get_dict_type(dct, max_typed_dict_size):
     """Return a TypedDict for `dct` if all the keys are strings.
     Else, default to the union of the keys and of the values."""
+    if len(dct) == 0:
+        # Special-case this because returning an empty TypedDict is
+        # unintuitive, especially when you've "disabled" TypedDict generation
+        # by setting `max_typed_dict_size` to 0.
+        return Dict[Any, Any]
```
carljm: 👍

monkeytype/stubs.py
pradeep90: Nothing :) I've made `stubs` a public attribute now.

monkeytype/stubs.py
pradeep90: Yeah, I'd have liked to use `Type[...]`. However, when I tried using just `cls: Type[List]` and left out the other possible container types, it typechecked even with `mypy --strict`. Not sure why it didn't catch the errors (probably because of the `type` annotations in `typing.pyi`).

I've left it as `type` as I'm not sure of the right type to use here.

PR #163, opened 2020-02-12
Add config to disable applying of TypedDict stubs.

pradeep90:
When `apply` for TypedDicts is disabled in the config, augment the type rewriter in the config with one that converts TypedDicts to Dicts. This means that, when TypedDicts are enabled, the traces will still be stored as TypedDicts but the types shrunk from the traces will be Dicts for `apply` and TypedDicts for other modes like `stub` or `diff`.

Note that TypedDict class stubs won't be generated for `apply` since that happens after the traced types are shrunk.

Convert nested TypedDicts correctly by using the above rewriter wherever we need to convert a TypedDicts to a Dict.


pradeep90:
Closing because we would rather implement `apply` for TypedDicts.

PR #168, opened 2020-03-16
Shrink non-total TypedDicts.

pradeep90:
Shrink TypedDicts by looking for fields that are present in all the TypedDicts and fields that are present only in some. Track required and non-required fields by adding a prefix to the keys.

When all fields are non-required, generate a non-total TypedDict class stub.
When there both required and non-required fields, generate a total base class and a non-total subclass.

Design choices:

+ If a key has different value types in the traced types, then we shrink those types to get its value type. This should not lead to large types because we shrink large union types to Any.

+ If the resulting TypedDict is larger than max_typed_dict_size, then fall back to `Dict[str, ...]`. This will convert all nested TypedDicts to Dicts as well. Add a rewriter for this.


monkeytype/stubs.py
```diff
@@ -542,13 +545,47 @@ def rewrite_TypedDict(self, typed_dict: type) -> type:
...
+                                        function_stubs=[],
+                                        attribute_stubs=attribute_stubs))
+            attribute_stubs = []
+            for name, typ in non_required_fields.items():
+                rewritten_type, stubs = self.rewrite_and_get_stubs(typ, class_name_hint=name)
+                attribute_stubs.append(AttributeStub(name, rewritten_type))
+                self.stubs.extend(stubs)
+            self.stubs.append(ClassStub(name=f'{subclass_name}({make_forward_ref(base_class_name)}, total=False)',
+                                        function_stubs=[],
+                                        attribute_stubs=attribute_stubs))
```
carljm: It seems this section has more duplicated code than necessary. Why can't we just `if required_fields` process those (and emit the total=True TypedDict), then `if non_required_fields` process those and emit the total=False one (possibly subclassing if we generated one with required fields).

monkeytype/typing.py
```diff
@@ -26,13 +27,42 @@
...
+    non_required_fields = non_required_fields or {}
+    assert all(not key.startswith(REQUIRED_FIELD_PREFIX)
+               and not key.startswith(NON_REQUIRED_FIELD_PREFIX)
+               for key in chain(required_fields.keys(), non_required_fields.keys()))
+
+    required_fields = {(REQUIRED_FIELD_PREFIX + key): value
+                       for key, value in required_fields.items()}
+    non_required_fields = {(NON_REQUIRED_FIELD_PREFIX + key): value
+                           for key, value in non_required_fields.items()}
+    return TypedDict(DUMMY_TYPED_DICT_NAME, {
```
carljm: I wonder if it would simplify this diff if we used our own MonkeyType-internal representation for `TypedDict` types, instead of using the actual `typing.TypedDict`. If we used an internal representation we could track required vs optional fields in a more explicit way (e.g. two separate dictionaries) instead of requiring all this string prefix stuff.

The main argument for using the actual `TypedDict` type is that it's more intuitive for people writing rewriters? But I think the presence of the key prefixes destroys that intuitiveness anyway; we might as well just use our own representation and document it.

monkeytype/typing.py
```diff
@@ -58,31 +88,47 @@ def __are_typed_dict_types_equal(type1: type, type2: type) -> bool:
...
+    Non-required fields are those that appear in only some of the TypedDicts.
+    If the same key has a different value type, then its value is the Union of the value types."""
+    num_typed_dicts = len(typed_dicts)
+    key_value_types_dict = defaultdict(list)
+    for typed_dict in typed_dicts:
+        for key, value_type in typed_dict.__annotations__.items():
+            # Assumption: An input TypedDict to shrink_types will have all
+            # fields as required, since it was constructed directly from a
+            # dictionary object.
+            assert key.startswith(REQUIRED_FIELD_PREFIX)
```
carljm: In this case, why not just skip the prefixes altogether in the TypedDicts constructed as input to this function?

(Though as mentioned above, I still think a custom class for internal representation of `TypedDict` may be a better approach.)

monkeytype/typing.py
```diff
@@ -58,31 +88,47 @@ def __are_typed_dict_types_equal(type1: type, type2: type) -> bool:
...
+                       for key, value_types in key_value_types_dict.items()
+                       if len(value_types) == num_typed_dicts}
+    non_required_fields = {key: shrink_types(tuple(value_types), max_typed_dict_size)
+                           for key, value_types in key_value_types_dict.items()
+                           if len(value_types) != num_typed_dicts}
+
+    if len(required_fields) + len(non_required_fields) > max_typed_dict_size:
+        value_types = (RewriteTypedDictToDict().rewrite(value_type)
+                       for value_type in chain(required_fields.values(), non_required_fields.values()))
+        return Dict[str, Union[tuple(value_types)]]
```
carljm: it seems like we could detect and handle this case before we've done all the work to `shrink_types` individually for each key.

monkeytype/typing.py
```diff
@@ -91,11 +137,8 @@ def shrink_types(types):
     if len(types) == 0:
         return Any
     if all(is_typed_dict(typ) for typ in types):
-        return shrink_typed_dict_types(types)
-    all_dict_types = tuple(typed_dict_to_dict(typ)
-                           if is_typed_dict(typ)
-                           else typ
-                           for typ in types)
+        return shrink_typed_dict_types(types, max_typed_dict_size)
+    all_dict_types = tuple(RewriteTypedDictToDict().rewrite(typ) for typ in types)
```
carljm: It seems like `RewriteTypedDictToDict` will do a recursive rewrite of all nested `TypedDict` to `Dict` also. But is that actually correct here? I don't think it is. All we've determined is that at the top level we don't have all typed dicts, so we need to convert them to regular dict. But why does that imply that any value type that is a `TypedDict` should also be converted to a regular dict?

monkeytype/typing.py
```diff
@@ -58,31 +88,47 @@ def __are_typed_dict_types_equal(type1: type, type2: type) -> bool:
...
+
+    required_fields = {key: shrink_types(tuple(value_types), max_typed_dict_size)
+                       for key, value_types in key_value_types_dict.items()
+                       if len(value_types) == num_typed_dicts}
+    non_required_fields = {key: shrink_types(tuple(value_types), max_typed_dict_size)
+                           for key, value_types in key_value_types_dict.items()
+                           if len(value_types) != num_typed_dicts}
+
+    if len(required_fields) + len(non_required_fields) > max_typed_dict_size:
+        value_types = (RewriteTypedDictToDict().rewrite(value_type)
```
carljm: Also here using a recursive rewriter seems wrong. Just because the top-level dict we are shrinking is too big to be represented as a `TypedDict` doesn't imply that all nested `TypedDict` should also be converted to `Dict`.

tests/test_stubs.py
```diff
@@ -659,6 +687,34 @@ def test_render_typed_dict_in_list(self):
...
+        expected = '\n'.join([
+            'from mypy_extensions import TypedDict',
+            '',
+            '',
+            f'class FooNonTotalTypedDict({make_forward_ref("FooTypedDict")}, total=False):',
+            '    b: str',
+            '',
+            '',
+            'class FooTypedDict(TypedDict):',
+            '    a: int',
```
carljm: Why do we choose to render these in this order, requiring the ForwardRef, when we could just render them in the other order instead?

monkeytype/stubs.py
pradeep90: Yeah, that was pretty awful duplication. I've extracted the common code. However, I'm keeping the branches so that it is explicit rather than implicit that one of them emits two stubs.

tests/test_stubs.py
pradeep90: We render them in alphabetical order. The only reason I went for the ForwardRef was because it seemed pretty brittle to have it work only because the base class name was < the subclass name. 

Let me know if you're ok with `FooTypedDict` and `FooTypedDictNonTotal` so that they are in alphabetical order.

monkeytype/typing.py
pradeep90: Yeah, my initial version was using `TypedDict({"required_fields": TypedDict(...), "non_required_fields": TypedDict(...)})`.

The main thing I wanted was to make the representation be a subtype of `type` so that it would be used seamlessly with the existing code. For example, `get_dict_type` would be able to return this `TypedDict` and we would be able to have `get_type([{"name": "foo"}])` be `List[TypedDict(...)]`, etc. Similarly, `shrink_typed_dict_types` would be able to return this `TypedDict` and we would be able to have recursive `TypedDicts`.

So, I've gone back to a `TypedDict` with one nested `TypedDict` each for required and optional fields. Let me know if that works for you. I tried defining a `class InternalTypedDict(type): ...` but I had trouble getting that to work and gave up.

monkeytype/typing.py
pradeep90: Yeah, I was trying to avoid sending `shrink_types` a TypedDict with optional fields because I didn't have a good idea of how to handle that case. Have fixed it and added tests :)

monkeytype/typing.py
pradeep90: I think that `shrink_types(AnonymousTypedDict({'a': AnonymousTypedDict({'b': int})}), Dict[str, int])` should be `Union[Dict[str, Dict[str, int]], Dict[str, int]]` rather than `Union[Dict[str, AnonymousTypedDict({'b': int})], Dict[str, int]]`. 

Given that a top-level mismatch probably means that the nested types won't match, it seems a bit weird to emit a TypedDict class for just one part of the Union. (This holds only for our anonymous TypedDicts. A user-defined TypedDict will be treated like any other type.)

monkeytype/typing.py
```diff
@@ -26,13 +27,33 @@
 
 
 DUMMY_TYPED_DICT_NAME = 'DUMMY_NAME'
+DUMMY_REQUIRED_TYPED_DICT_NAME = 'REQUIRED_TYPED_DICT_NAME'
+DUMMY_OPTIONAL_TYPED_DICT_NAME = 'OPTIONAL_TYPED_DICT_NAME'
+REQUIRED_FIELD_PREFIX = 'REQUIRED__'
+OPTIONAL_FIELD_PREFIX = 'OPTIONAL__'
```
carljm: It seems these are only used in `remove_requiredness_prefix`, and that function isn't used anywhere? Just remove?

monkeytype/typing.py
carljm: Sounds good!

tests/test_stubs.py
carljm: Yeah that seems fine.

carljm:
Oh also it seems the tests are failing on 3.6.

pradeep90:
I've kept `types_equal` in typing.py because I got an import error when putting it in compat.py:

```
monkeytype/compat.py:43:5: error: Module 'mypy_extensions' has no attribute '_TypedDictMeta'; maybe "_TypedDict" or "TypedDict"?
```

Also, I'm pretty unhappy with the `types_equal` function :| . It can't handle the following case:

```
(List[make_typed_dict(required_fields={'a': List[make_typed_dict(required_fields={'a': int})]})],
 List[make_typed_dict(required_fields={'a': List[make_typed_dict(required_fields={'a': int})]})],
 True),
```

because we'd have to override `__eq__` for dict when comparing the annotations of a `TypedDict`. So, currently, 3.6 and 3.7 diverge for this kind of case. Would be great if you can suggest a better alternative, maybe in the future.

monkeytype/typing.py
```diff
@@ -26,13 +27,55 @@
...
+DUMMY_OPTIONAL_TYPED_DICT_NAME = 'OPTIONAL_TYPED_DICT_NAME'
 
 
 # Functions like shrink_types and get_type construct new types at runtime.
 # Mypy cannot currently type these functions, so the type signatures for this
 # file live in typing.pyi.
 
 
+def types_equal(typ, other_type) -> bool:
+    # Types for which equality with inner TypedDict doesn't work correctly on 3.6.
```
carljm: The comment here suggests this is only necessary for 3.6, but it seems like we use this function unconditionally on any Python version? The reason I suggested putting this in `compat.py` is that I thought we could isolate the hackiness to 3.6 and maybe use normal equality checks on 3.7?

carljm:
It seems like it's not hard for `types_equal` to handle the case you outline, it just requires a bit more special-casing of TypedDict in that function (to actually compare the full annotations item by item and bypass dict equality).

And I'm not sure what to take from the import error from `compat.py`, I'd have to see the full code of `compat.py` and what line that error occurs on. What is trying to import `_TypedDictMeta` from `mypy_extensions`?

monkeytype/typing.py
pradeep90: Yeah, that's what I was going for, but got stymied by the import error.

Turns out that all imports in compat.py had the error and had needed `type: ignore` :)

pradeep90: The best solution would be to override `List.__eq__`, just like we do for `TypedDict`, but that didn't work in the REPL when I tried it. I'm going to leave it as is for now, since we use it in only one place.

carljm:
Awesome work!! Last thing: can you add an entry in `CHANGES.rst` for this change? Then I will merge.

carljm:
Thank you!! Merged.

PR #172, opened 2020-04-01
Use libcst to apply annotations instead of retype.

### LINKED ISSUE #76: Special-cases in `render_annotations` are not recursively applied in generics
In our `render_annotations` function, we defer to the `__repr__` methods in the `typing` module for rendering generic types in annotations. But this means that the _other_ special cases in `render_annotations` are not applied to types within a generic. For a simple example, consider the type `Dict[str, Optional[int]`. We will render this as `Dict[str, Union[int, NoneType]]` instead, because neither our special case for `Optional` nor our special case for `NoneType -> None` will be applied within the `Dict`.

Some of our other special cases (e.g. for `NewType` or for `builtins`) are also implemented in some form by the `typing` module itself, so they don't appear to break inside a generic.

But basically we have two versions of type-repr, one inside `typing`, and one in our own `render_annotations`; the latter applies only to top level types, the former to everything inside a generic. So we can't add new cases to `render_annotations` and expect them to apply universally. These should be unified in some form, though I'm not sure exactly how to do so; we might need to do our own recursive-descent rendering for generics (though there's probably some private methods/functions in `typing` we can use to avoid reimplementing too much).

pradeep90:
+ This enables `monkeytype apply` for our generated TypedDict classes. Add sanity tests.

    I've changed the ExistingAnnotationStrategy for `apply` to `REPLICATE` because the ApplyTypeAnnotationsVisitor doesn't apply function stubs that don't have a return type. (I'll add a flag and feature for this upstream in libcst, but this strategy should be ok, since I'm guessing `OMIT` was used mainly to please `retype`.)

    NOTE: libcst didn't work on 3.6 because it uses dataclasses. I've added a backport of that package as per https://pypi.org/project/dataclasses/.

+ Render annotations recursively. I noticed this after `apply` was fixed :D. Things like `List[MovieTypedDict]` showed up as `List[ForwardRef('MovieTypedDict')]`. Fixes #76. 

    Please see the commit message for more details.

+ Append __RENAME_ME__ to the generated TypedDict names so that it is clearer that people have to rename them. Plus, `d: DTypedDict` looked weird.


tests/test_cli.py
```diff
@@ -364,3 +346,76 @@ def my_test_function(a, b):
                 ret = cli.main(['apply', 'my_test_module'], stdout, stderr)
     assert ret == 0
     assert 'warning:' not in stdout.getvalue()
+
+
+def test_apply_stub_using_libcst():
+    source = """
```
carljm, line 352: nit: can we use `textwrap.dedent()` to avoid having all this code at top-level indent?

monkeytype/stubs.py
```diff
@@ -298,31 +299,64 @@ def _get_optional_elem(anno: Any) -> Any:
...
-        rendered = repr(anno.__forward_arg__)
-    elif getattr(anno, '__module__', None) == 'typing':
-        rendered = repr(anno).replace('typing.', '')
-    elif isinstance(anno, NoneType):
-        rendered = 'None'
-    elif isinstance(anno, type):
-        if anno.__module__ in ('builtins',):
-            rendered = anno.__qualname__
+class RenderAnnotation(GenericTypeRewriter[str]):
+    """Render annotation recursively."""
```
carljm, line 303: This is awesome! You just fixed another significant longstanding bug in passing :)

pradeep90:
Small fixes: Used `self.make_builtin_typed_dict()` instead of directly using `TypedDict()` in the base class. Added `self.rewrite_malformed_container` for the base case in `_rewrite_container` instead of returning the container itself.

PR #175, opened 2020-04-16
Remove the tilde prefix from rendered type variables.

pradeep90:
They were being rendered as `Dict[str, ~T]` because str(T) = '~T'.


carljm:
Thank you!

PR #177, opened 2020-04-18
Fix an example of CallTracer

thombashi:
An example of `CallTracer` seems does not reflect newly added positional argument (missing `max_typed_dict_size`)

https://github.com/Instagram/MonkeyType/blob/56a688efccedd66521542144cd0afa4dc1e5e562/monkeytype/tracing.py#L190-L196


carljm:
Thank you!

PR #183, opened 2020-05-14
Allow `apply` to overwrite existing annotations using CLI flag.

pradeep90:
It should be safe to remove the `args.existing_annotation_strategy = ExistingAnnotationStrategy.REPLICATE` line from `apply_stub_handler` because that is now the default value in `args`.

This needed the latest ApplyTypeAnnotationsVisitor from libcst.  Set the minimum libcst version to 0.3.5. (Another reason for this PR was that I unthinkingly changed a public function signature in a backward-incompatible way once again :(. Hope to keep that in mind next time.)

Note: I ran `pipenv update -d` to update the packages. This ended up upgrading pyflakes, which [now errors](https://github.com/PyCQA/pyflakes/pull/516) on `List[make_forward_ref('Foo')]` saying "undefined name 'Foo'", probably because it's a literal non-class string within a container. So, I've temporarily silenced these errors in .flake8. Lmk if you have a good solution for that other than creating a bug report there.

Fixed f-strings that didn't have expressions inside.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,2 +1,5 @@
 [flake8]
 max-line-length = 120
+per-file-ignores =
+    tests/test_stubs.py:F821
+    tests/test_typing.py:F821
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -24,10 +24,10 @@
         },
         "appdirs": {
             "hashes": [
-                "sha256:9e5896d1372858f8dd3344faf4e5014d21849c756c8d5701f78f8a103b372d92",
-                "sha256:d8b24664561d0d34ddfaec54636d502d7cea6e29c3eaf68f3df6180863e2166e"
+                "sha256:7d5d0167b2b1ba821647616af46a749d1c653740dd0d2415100fe26e27afdf41",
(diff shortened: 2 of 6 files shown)


.flake8
```diff
@@ -1,2 +1,5 @@
 [flake8]
 max-line-length = 120
+per-file-ignores =
+    tests/test_stubs.py:F821
+    tests/test_typing.py:F821
```
carljm, line 5: why should this lint rule be ignored in these files?

carljm:
Looks good! Just curious about the flake8 change.

.flake8
pradeep90: I'd mentioned that in the summary:

> Note: I ran pipenv update -d to update the packages. This ended up upgrading pyflakes, which now errors on List[make_forward_ref('Foo')] saying "undefined name 'Foo'", probably because it's a literal non-class string within a container. So, I've temporarily silenced these errors in .flake8. Lmk if you have a good solution for that other than creating a bug report there.

carljm: Ah, sorry for missing the summary! Yeah that's a lot of lint errors, I don't have a better solution.

PR #201, opened 2020-06-26
MonkeyType: require a newer version of LibCST

### LINKED ISSUE #198: monkeytype destroys default values of keyword-only args
Here's a repro:
```
(.env) ~/tmp/bug λ cat mod.py   
def foo(a, *, b=5):
    return (a, b)


foo("asdf", b=10)
(.env) ~/tmp/bug λ cat script.py
from mod import *
(.env) ~/tmp/bug λ monkeytype run script.py
(.env) ~/tmp/bug λ monkeytype apply mod    
from typing import Tuple

def foo(a: str, *, b = ...) -> Tuple[str, int]:
    return (a, b)


foo("asdf", b=10)

(.env) ~/tmp/bug λ cat mod.py   
from typing import Tuple

def foo(a: str, *, b = ...) -> Tuple[str, int]:
    return (a, b)


foo("asdf", b=10)
```

hauntsaninja:
Fixes #198

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ setup(
         ]
     },
     python_requires='>=3.6',
-    install_requires=['mypy_extensions', 'libcst>=0.3.5'],
+    install_requires=['mypy_extensions', 'libcst>=0.3.7'],
     classifiers=[
         'Development Status :: 5 - Production/Stable',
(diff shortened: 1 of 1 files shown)


carljm:
Thank you! ✨

PR #218, opened 2020-12-03
Add Pyre configuration for local and CI typechecking.

pradeep90:
Request for comments!

Feature-wise, there don't seem to be any issues in migrating from Mypy to Pyre. However, Pyre's open-source integration isn't all that mature. I can understand if people object to this :)

The main wrinkle is that Pyre needs to know the search path and typeshed path. We can't hardcode them in the configuration because they would be relative to the virtual environment. Plus, CircleCI would have a different location from the developer. (This is one of the main pain points of open-source Pyre usage.) 

So,  I'm using a similar approach to that of [Libcst](https://github.com/Instagram/LibCST/blob/master/.circleci/config.yml):

+ We will have a template in `.pyre_configuration.example`.
+ The CircleCI script will copy over that template and specify `--search-path` and `--typeshed` on the command-line. 
+ Developers would also copy over the template to `.pyre_configuration` and hardcode their search path locally. Given that it is git-ignored, those changes would remain local. Alternatively, they too could use the template as is and specify the paths via flags.

Need to test and see if CircleCI works. Open to better suggestions.

The 47 new type errors seem pretty reasonable. Once we can agree on the overall migration from Mypy to Pyre, agree on a good approach, and get CircleCI working, I'll patch or ignore each error.

- [ ]  Use strict mode.
- [ ] Update docs. Developers would have to install `watchman`.
- [ ] Remove Mypy, which is currently needed by the `demo/`.


carljm (changes requested):
As discussed offline, we'd like to get to contributor workflow parity before making this switch.

carljm:
Closing for now due to lack of activity; can reopen if we want to look into this again; I know Pyre OSS usage has improved recently.

PR #226, opened 2021-03-08
Update `requirements.txt`

ansley:
I wasn't able to run the MonkeyType demos on my machine because of a missing header file in `typed_ast`. This is a [known issue](https://github.com/python/typed_ast/issues/97) that can be fixed with an updated version of the package. (I'm using CPython, and it was still an issue for me, even though the original poster is using PyPy.) Then, to support the new version of `typed_ast`, I  [updated `mypy`](https://github.com/python/typed_ast/issues/124). Finally, I fixed some additional warnings by updating other packages.

Test plan: The "Quickstart" guide in `demo/README.md` can be completed successfully

### DIFF
--- a/demo/requirements.txt
+++ b/demo/requirements.txt
@@ -1,4 +1,4 @@
-attrs==17.4.0
+attrs==19.2.0
 black==18.4a2
 click==6.7
 coverage==4.5.1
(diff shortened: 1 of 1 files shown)


carljm:
Thanks!

PR #249, opened 2021-10-03
Add caller column into stub

ulwlu:
## What this pr does

- add caller column into stub

## Why I do this

If you have a code like this.

```python
# File: main.py
import a
import b


# File: a.py
from test import return_str
print(return_str(15, 10))


# File: b.py
from test import return_str
print(return_str("a", "b"))


# File: test.py
def return_str(a, b):  # return_str expects only accepting str.
    return a, b
```

This generates stub like below.

```tsv
created_at	module	qualname	arg_types	return_type	yield_type
2021-10-03 15:51:45.214250	test	return_str	{"a": {"module": "builtins", "qualname": "int"}, "b": {"module": "builtins", "qualname": "int"}}	{"elem_types": [{"module": "builtins", "qualname": "int"}, {"module": "builtins", "qualname": "int"}], "module": "typing", "qualname": "Tuple"}
2021-10-03 15:51:45.214287	test	return_str	{"a": {"module": "builtins", "qualname": "str"}, "b": {"module": "builtins", "qualname": "str"}}	{"elem_types": [{"module": "builtins", "qualname": "str"}, {"module": "builtins", "qualname": "str"}], "module": "typing", "qualname": "Tuple"}
```

Then you'll notice there is some frame is calling return_str with int arguments which is wrong. You want to find where this calls, but you can't see any information about where that is. So you need to delete the wrong stub, apply the stub, run mypy and find where that happens.

By this PR, you can find where to fix by once.

```tsv
created_at	module	qualname	arg_types	return_type	yield_type	caller
2021-10-03 15:51:45.214250	test	tmp	{"a": {"module": "builtins", "qualname": "int"}, "b": {"module": "builtins", "qualname": "int"}}	{"elem_types": [{"module": "builtins", "qualname": "int"}, {"module": "builtins", "qualname": "int"}], "module": "typing", "qualname": "Tuple"}		${filename}/a.py:${module-name(if you run in top module, this doesn't show up.):}2
2021-10-03 15:51:45.214287	test	tmp	{"a": {"module": "builtins", "qualname": "str"}, "b": {"module": "builtins", "qualname": "str"}}	{"elem_types": [{"module": "builtins", "qualname": "str"}, {"module": "builtins", "qualname": "str"}], "module": "typing", "qualname": "Tuple"}		${filename}/b.py:${module-name(if you run in top module, this doesn't show up.):}2

```

As you see calling function with `(int, int)` arguments is happened at `"a.py:2"`.

## remarks

I submitted Contributor License Agreement.

### DIFF
--- a/doc/stores.rst
+++ b/doc/stores.rst
@@ -154,6 +154,11 @@ choose to implement its own alternative (de)serialization.
     A JSON-serialized representation of the actual yield type for this traced
     call, or ``None`` if this call did not yield (i.e. returned instead).
 
+  .. attribute:: caller: Optional[type]
+
+    The ``caller`` of the traced function or method contains filename, module, and line number information.
+
+
 .. currentmodule:: monkeytype.db.base
 
 CallTraceThunk
--- a/doc/tracing.rst
+++ b/doc/tracing.rst
@@ -178,7 +178,7 @@ method saves all collected traces to the given ``store``.
 CallTrace
 '''''''''
 
-.. class:: CallTrace(func: Callable, arg_types: Dict[str, type], return_type: Optional[type] = None, yield_type: Optional[type] = None)
+.. class:: CallTrace(func: Callable, arg_types: Dict[str, type], return_type: Optional[type] = None, yield_type: Optional[type] = None, caller: Optional[str] = None)
 
   Type information for one traced call of one function.
 
@@ -202,3 +202,7 @@ CallTrace
   .. attribute:: yield_type: Optional[type]
 
     Type yielded by this call, or ``None`` if this call did not yield.
+
+  .. attribute:: caller: Optional[type]
+
+    The filename, module, and line number information that executes function where the trace occurred.
--- a/monkeytype/db/sqlite.py
+++ b/monkeytype/db/sqlite.py
@@ -39,7 +39,8 @@ CREATE TABLE IF NOT EXISTS {table} (
   qualname    TEXT,
   arg_types   TEXT,
   return_type TEXT,
-  yield_type  TEXT);
+  yield_type  TEXT,
+  caller      TEXT);
 """.format(table=table)
     with conn:
         conn.execute(query)
@@ -52,7 +53,7 @@ ParameterizedQuery = Tuple[str, List[QueryValue]]
 def make_query(table: str, module: str, qualname: Optional[str], limit: int) -> ParameterizedQuery:
     raw_query = """
     SELECT
-        module, qualname, arg_types, return_type, yield_type
+        module, qualname, arg_types, return_type, yield_type, caller
     FROM {table}
     WHERE
         module == ?
@@ -63,7 +64,7 @@ def make_query(table: str, module: str, qualname: Optional[str], limit: int) -> 
         values.append(qualname)
     raw_query += """
     GROUP BY
-        module, qualname, arg_types, return_type, yield_type
+        module, qualname, arg_types, return_type, yield_type, caller
     ORDER BY date(created_at) DESC
     LIMIT ?
     """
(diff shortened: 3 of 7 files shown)


doc/tracing.rst
```diff
@@ -202,3 +202,7 @@ CallTrace
   .. attribute:: yield_type: Optional[type]
 
     Type yielded by this call, or ``None`` if this call did not yield.
+
+  .. attribute:: caller: Optional[type]
```
carljm, line 206: I think this should be `Optional[str]` not `Optional[type]`?

doc/stores.rst
```diff
@@ -154,6 +154,11 @@ choose to implement its own alternative (de)serialization.
     A JSON-serialized representation of the actual yield type for this traced
     call, or ``None`` if this call did not yield (i.e. returned instead).
 
+  .. attribute:: caller: Optional[type]
```
carljm, line 157: Should be `Optional[str]` not `Optional[type]`?

monkeytype/db/sqlite.py
```diff
@@ -63,7 +64,7 @@ def make_query(table: str, module: str, qualname: Optional[str], limit: int) ->
         values.append(qualname)
     raw_query += """
     GROUP BY
-        module, qualname, arg_types, return_type, yield_type
+        module, qualname, arg_types, return_type, yield_type, caller
```
carljm, line 67: I don't think we need to group by caller here?

monkeytype/encoding.py
```diff
@@ -195,13 +195,15 @@ def __init__(
         qualname: str,
         arg_types: str,
         return_type: Optional[str],
-        yield_type: Optional[str]
+        yield_type: Optional[str],
+        caller: Optional[str]
```
carljm, line 199: Can we make this default to `None` for better backward-compatibility with existing third-party extensions of MonkeyType that may be instantiating `CallTraceRow`?

monkeytype/tracing.py
```diff
@@ -161,6 +165,18 @@ def get_func(frame: FrameType) -> Optional[Callable]:
     return func
 
 
+def get_caller_func_info(frame: FrameType) -> Optional[str]:
+    """Return the information of function who calls func who has code object."""
+    caller_frame = frame.f_back
+    if caller_frame is None:
+        return None
```
carljm, line 172: It seems unnecessary to have these lines duplicated both here and in `_get_caller_func_info` method below. I would instead have this function be called `get_frame_location_info` and have it take the caller frame instead of the called frame, i.e. just eliminate these three lines from it.

tests/test_tracing.py
```diff
@@ -89,6 +99,39 @@ def test_get_func(self, frame, expected_func):
...
+        ],
+    )
+    def test_get_caller_func_info(self, frame):
+        caller_frame = frame.f_back
+        if caller_frame is None:
+            return None
+        caller_code = caller_frame.f_code
+        filename = caller_code.co_filename
+        funcname = caller_code.co_name if caller_code.co_name != "<module>" else None
+        lineno = str(caller_frame.f_lineno)
```
carljm, line 121: duplicating the code in the test doesn't test the correctness of the code, it just tests that you copy-pasted correctly :) instead please put the expected final string into the test parameters for each case above, and check against that. This may require using a placeholder for the initial part of the path which can vary depending on where the tests are run, and then replacing that placeholder inside the test with the actual path for the current test run.

tests/test_tracing.py
```diff
@@ -184,7 +227,16 @@ class TestTraceCalls:
...
         with trace_calls(collector, max_typed_dict_size=0):
             simple_add(1, 2)
-        assert collector.traces == [CallTrace(simple_add, {'a': int, 'b': int}, int)]
+        assert collector.traces == [
+            CallTrace(
+                simple_add,
+                {"a": int, "b": int},
+                int,
+                caller=pytestRegex(
+                    f"^{get_cur_filename()}:{get_cur_funcname()}:[0-9]*$"
```
carljm, line 236: this is pretty verbose and repetitive in all these tests. I think we can improve it a lot by defining a wrapper or subclass of `pytestRegex` called like `match_caller` that always adds to the pattern the initial current-filename (which can just be based on `__file__`, you don't need to use frame-walking for this) and the final line number match part, so each test can just provide only the function name, say something like `caller=match_caller("TestTraceCalls.test_simple_call")` (I think I'd also prefer just hardcoding the function name for simplicity rather than using `get_cur_funcname()`)

carljm (changes requested):
This seems like useful data to have in the MonkeyType database, even if it isn't reflected in the typing stub output; thanks for the contribution!

Some comments before it's merge-ready.

tests/test_tracing.py
ulwlu: Sorry for low quality of test codes, I understand. I'll fix.

Issue #257, opened 2022-03-23
Formatter?


eggplants:
I suggest to introduce [black](https://github.com/psf/black) like [libcst](https://github.com/Instagram/MonkeyType/blob/cef09c6702fbab58e66582f76f2559d6a21e448c/demo/requirements.txt#L2) and [isort](https://github.com/pycqa/isort).

eggplants:
>Not sure how libcst is relevant here

When I thought about what formatter to adopt, I found that libcst uses black and suggested it.

https://github.com/Instagram/MonkeyType/blob/cef09c6702fbab58e66582f76f2559d6a21e448c/demo/requirements.txt#L2

carljm:
Sorry, my reading comprehension fail -- you said "black _like_ libcst", I read it as a "black and libcst and isort" :)

PR #258, opened 2022-03-23
Introduce setup.cfg and black+isort

eggplants:
Close: #257


setup.cfg
```diff
@@ -0,0 +1,95 @@
...
+extend-ignore =
+    # E203 whitespace before ':'
+    # https://github.com/psf/black/issues/315
+    E203
+per-file-ignores =
+    tests/test_stubs.py:F821
+    tests/test_typing.py:F821
+
+[mypy]
+python_version = 3.6
```
carljm, line 72: this should also be 3.7

carljm (approved):
Looks great! Thank you for doing this! Just should bump the min Python version up to 3.7 a couple places, but I'd also be fine if that happens in a separate PR.

setup.cfg
eggplants: da80482

setup.cfg
```diff
@@ -0,0 +1,95 @@
...
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
+    Programming Language :: Python :: 3.9
+
+[options]
+packages = find:
+include-package-data = True
+python_requires = >= 3.6
```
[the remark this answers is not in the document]
eggplants: da80482

PR #259, opened 2022-03-23
Make mypy config more strict

eggplants:
`mypy --strict`


carljm:
Looks good! Can you rebase on the commit I pushed a bit ago to get the build green? (It just ignored one new mypy error that showed up after the bump to python version 3.7.)

carljm:
Thank you!

Issue #256, opened 2022-03-13
Add `--all` to `monkeytype stub` + `monkeytype apply`


eggplants:
I always use:

```
pip install pytest-monkeytype
pytest --monkeytype-output=./monkeytype.sqlite3
for i in $(monkeytype list-modules);do echo "[$i]";monkeytype stub $i;done
for i in $(monkeytype list-modules);do echo "[$i]";monkeytype apply $i;done
...
```

I would like to add the `--all` option to run the command against all of the output modules of `monkeytype list-modules`.

Usage: `monkeytype stub --all && monkeytype apply --all`

carljm:
See #26 for design discussion of this feature. It's in a "pull requests welcome" state :)

PR #261, opened 2022-04-01
Add `stub-all` and `apply-all`

eggplants:
Closes: #26 (+#256)

---

I have implemented two sub commands, `stub-all` and `apply-all`.
It enables to generate and apply stubs of all modules listed on `list-modules` at once.

Try: `monkeytype stub-all` and `monkeytype apply-all`


eggplants:
(No test yet...)

carljm:
Hi @eggplants ! Sorry for slow review here, somehow I missed this PR.

As I mentioned on #26, I think the better approach here would be a `-r` flag to apply stubs recursively to a package and all its submodules. I prefer this because it is more controlled/explicit and less likely to do unexpected things if you accidentally have a bunch of extra stuff in your trace store.

I also don't think we need to provide this for `stub`.

PR #269, opened 2022-06-14
Bump required libcst version

zsol:
LibCST has seen a bunch of important fixes to apply type annotation and others that affect MonkeyType. For example https://github.com/Instagram/LibCST/issues/690 should be fixed now.

Note to reviewers: LibCST 0.4.0 and onwards comes with binary wheels, whereas before it was a pure python package.


carljm:
Thank you!

Issue #203, opened 2020-07-21
add `from __future__ import annotations` if adding forward-reference annotations


Jongy:
```
$ cat a.py
class A:
    def f(self):
        return self

$ cat b.py
import a

a.A().f()

$ monkeytype run b.py
$ monkeytype apply a
class A:
    def f(self) -> A:
        return self

$ python b.py
Traceback (most recent call last):
  File "b.py", line 1, in <module>
    import a
  File "a.py", line 1, in <module>
    class A:
  File "a.py", line 2, in A
    def f(self) -> A:
NameError: name 'A' is not defined
```

As described [here](https://www.python.org/dev/peps/pep-0484/#forward-references), I guess forward declarations should have been used here. If I manually use `-> "A"`, then Mypy identifies it correctly.

I see references to `typing.ForwardRef` in MonkeyType's code, so either it was missed for some reason, or am I using it wrong?

MonkeyType version: `20.5.0`.

carljm:
Thanks for the report! The `ForwardRef` references in MonkeyType's code are for cases where we get the existing annotation from the code and need to reproduce it in our stub. MonkeyType doesn't currently attempt to detect forward references in generated stubs and output stringified references. And I don't think it makes sense to add that feature at this point, because it's unnecessary with `from __future__ import annotations`, which in future will be the default behavior of annotations.

So if you aren't using `from __future__ import annotations`, this is just something you need to fix up as needed in the applied annotations. In general MonkeyType doesn't aim to provide ready-for-production code; it aims to provide an informative first draft of type information to be edited as needed by the developer. This also applies to things like generating concrete types when perhaps an abstract type would be more suitable, or unions where a common parent type would be better.

Jongy:
Thanks for the quick reply! I wasn't aware of `from __future__ import annotations`. Like you suggested, I will use it. 

If MonkeyType doesn't attempt to detect forward references at all, perhaps `monkeytype apply` should also insert `from __future__ import annotations`? (just like it inserts any missing import from `typing`)

carljm:
Yeah, that's a reasonable thought. I'm a little hesitant to have MonkeyType do that by default, though, since `from __future__ import annotations` can be a breaking change (if your code is inspecting the annotations at runtime).

Jongy:
Hmm. Then maybe MonkeyType can identify it is about to add a forward reference (if it's easy given the current code) and only if so, inject `from __future__ import annotations`. If done only when required, it is much less of a breaking change.

[...]

Issue #111, opened 2018-09-06
Monkeytype creates circular imports


bluebird75:
With MonkeyType 18.8.0 and Python 3.6 .

I ran into this on a complex example, but it's actually quite easy to reproduce.

```
d:\work\pyann\monkeytype\toto>type a.py
class A:
    def __init__(self):
        self.titi = 'A_titi'

    def copy_from(self, other):
        self.titi = other.titi

d:\work\pyann\monkeytype\toto>type b.py
from a import A

class B(A):
    def __init__(self):
        super().__init__()
        self.titi = 'B_titi'

d:\work\pyann\monkeytype\toto>type main.py
from a import A
from b import B

if __name__ == '__main__':
    a = A()
    print( a.titi )
    b = B()
    a.copy_from( b )
    print( a.titi )

d:\work\pyann\monkeytype\toto>python main.py
A_titi
B_titi

d:\work\pyann\monkeytype\toto>monkeytype run main.py
A_titi
B_titi

d:\work\pyann\monkeytype\toto>monkeytype list-modules
a
b

d:\work\pyann\monkeytype\toto>monkeytype apply a

d:\work\pyann\monkeytype\toto>monkeytype apply b
Traceback (most recent call last):
  File "c:\program files (x86)\python36-32\lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "c:\program files (x86)\python36-32\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "C:\Program Files (x86)\Python36-32\Scripts\monkeytype.exe\__main__.py", line 9, in <module>
  File "c:\program files (x86)\python36-32\lib\site-packages\monkeytype\cli.py", line 377, in entry_point_main
    sys.exit(main(sys.argv[1:], sys.stdout, sys.stderr))
  File "c:\program files (x86)\python36-32\lib\site-packages\monkeytype\cli.py", line 362, in main
    handler(args, stdout, stderr)
  File "c:\program files (x86)\python36-32\lib\site-packages\monkeytype\cli.py", line 130, in apply_stub_handler
    stub = get_stub(args, stdout, stderr)
  File "c:\program files (x86)\python36-32\lib\site-packages\monkeytype\cli.py", line 105, in get_stub
    traces.append(thunk.to_trace())
  File "c:\program files (x86)\python36-32\lib\site-packages\monkeytype\encoding.py", line 183, in to_trace
[...]

PR #282, opened 2022-12-25
Move new imports to if TYPE_CHECKING block

sagarbadiyani:
Fix for https://github.com/Instagram/MonkeyType/issues/111 and https://github.com/Instagram/MonkeyType/issues/203

### DIFF
--- a/tests/test_type_checking_imports_transformer.py
+++ b/tests/test_type_checking_imports_transformer.py
@@ -1,0 +1,158 @@
+from typing import Type, List
+
+from libcst.codemod import CodemodTest, Codemod, CodemodContext
+from libcst.codemod.visitors import ImportItem
+
+from monkeytype.type_checking_imports_transformer import MoveImportsToTypeCheckingBlockVisitor
+
+
+class TestMoveImportsToTypeCheckingBlockVisitor(CodemodTest):
+
+    TRANSFORM: Type[Codemod] = MoveImportsToTypeCheckingBlockVisitor
+
+    def run_test_case(
+        self,
+        import_items_to_be_moved: List[ImportItem],
+        before: str,
+        after: str,
+    ) -> None:
+        context = CodemodContext()
+        MoveImportsToTypeCheckingBlockVisitor.store_imports_in_context(
+            context,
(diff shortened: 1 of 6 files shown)


sagarbadiyani:
@carljm can you please take a look?

carljm (changes requested):
Hi, thanks for the contribution! The code looks OK, but I think we need some tests for this new functionality, and an update to the docs including the new flag.

sagarbadiyani:
Added a few test cases and updated the docs @carljm, can you please take a look again?

tests/test_type_checking_imports_transformer.py
```diff
@@ -0,0 +1,162 @@
...
+            from __future__ import annotations
+            
+            from typing import TYPE_CHECKING
+            
+            if TYPE_CHECKING:
+                import c.C
+                from a import B
+            
+            if TYPE_CHECKING:
+                from d import E
```
carljm, line 76: It would be nice if in this case we could find and append to the existing `if TYPE_CHECKING:` block rather than adding a new one. It seems like in principle that should be doable, but I'm assuming you looked at it and felt it was too complex? It could also always be done as a follow-up improvement if anyone cares enough.

carljm (changes requested):
This looks great! Thanks for adding the tests and docs.

I think this is basically merge-ready, but it looks like it is failing CI on the flake8 linter check -- a lot of blank lines were added with whitespace in them, which the linter doesn't like.

Sorry for the inconvenience that GitHub won't run CI automatically when you push, since you're a first-time contributor I have to trigger it to run. But all the checks that the CI does should also be runnable locally via `tox`.

tests/test_type_checking_imports_transformer.py
sagarbadiyani: Correct, it should be do-able but it was getting too complicated, that's why took the debt, will try to fix this in the subsequent PR :)

sagarbadiyani:
Fixed the failing `flake8` check @carljm and ran `tox` locally
Thank you for your review

carljm (approved):
Thank you for adding this feature!

PR #291, opened 2023-05-23
Upgrade pip dependencies

mengdilin:
This is a generated change running the commands `pipenv shell && pipenv upgrade requests==2.31.0` to resolve a vulnerability in the requests dependency

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -20,3 +20,6 @@ tox = "*"
 setuptools_rust = "*"
 libcst = "*"
 importlib_metadata = "*"
+
+[packages]
+requests = "==2.31.0"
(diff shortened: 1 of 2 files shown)


carljm:
Thanks for the PR! I don't think this is quite right, because a) `requests` is just a dependency of one of our dev dependencies, it's not a runtime dependency of MonkeyType, so we don't want to add it as a dependency in `[packages]`, and b) we don't want to pin the `requests` version exactly, as that will prevent it from being updated in future. We could set a minimum version for it, but really I think all we need to do here is update the lockfile without any changes to the source `Pipfile`; that should get us a new enough version of `requests`.

I can take this from here if you want.

carljm:
Fixed in #292

Issue #300, opened 2023-08-25
Adding an index to monkeytype.sqlite3 speeds up stub generation


Feuermurmel:
I've noticed that adding an index to `monkeytype.sqlite3` speeds up stub generation a bit:

```
(venv) $ time monkeytype stub foo.bar > /dev/null

real	0m1.673s
user	0m1.242s
sys	0m0.419s
(venv) $ sqlite3 monkeytype.sqlite3
SQLite version 3.37.0 2021-12-09 01:34:53
Enter ".help" for usage hints.
sqlite> create index foo on monkeytype_call_traces (module, created_at);
sqlite> ^D
(venv) $ time monkeytype stub foo.bar > /dev/null

real	0m1.009s
user	0m0.873s
sys	0m0.104s
```

My `monkeytype.sqlite3` is 2.6 GB in size, which resulted from running our test suite once.

carljm:
Nice find! Would you be up for submitting a PR that adds this index to initialization of the sqlite db?

PR #301, opened 2023-08-26
Add index to call trace DB to speed up lookup by module

Feuermurmel:
Resolves https://github.com/Instagram/MonkeyType/issues/300.

### DIFF
--- a/monkeytype/db/sqlite.py
+++ b/monkeytype/db/sqlite.py
@@ -21,19 +21,26 @@ DEFAULT_TABLE = "monkeytype_call_traces"
 def create_call_trace_table(
     conn: sqlite3.Connection, table: str = DEFAULT_TABLE
 ) -> None:
-    query = """
-CREATE TABLE IF NOT EXISTS {table} (
-  created_at  TEXT,
-  module      TEXT,
-  qualname    TEXT,
-  arg_types   TEXT,
-  return_type TEXT,
-  yield_type  TEXT);
-""".format(
-        table=table
-    )
+    queries = [
+        """
+        CREATE TABLE IF NOT EXISTS {table} (
+          created_at  TEXT,
+          module      TEXT,
+          qualname    TEXT,
+          arg_types   TEXT,
+          return_type TEXT,
+          yield_type  TEXT);
+        """,
+        """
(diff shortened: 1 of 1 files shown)


monkeytype/db/sqlite.py
```diff
@@ -21,19 +21,28 @@
...
+        CREATE INDEX IF NOT EXISTS {table}_module ON {table} (module);
+        """
+    ]
+
     with conn:
-        conn.execute(query)
+        for query in queries:
+            conn.execute(
+                query.format(table=table)
+            )
```
carljm, line 45:
```suggestion
            conn.execute(query.format(table=table))
```

monkeytype/db/sqlite.py
```diff
@@ -21,19 +21,28 @@
...
+          qualname    TEXT,
+          arg_types   TEXT,
+          return_type TEXT,
+          yield_type  TEXT);
+        """,
+        """
+        -- This index speeds up lookups of call traces of a single module
+        -- (see `make_query()`).
+        CREATE INDEX IF NOT EXISTS {table}_module ON {table} (module);
+        """
```
carljm, line 38:
```suggestion
        """,
```

carljm:
Hi @Feuermurmel, this looks great! It needs a couple minor tweaks (per my comments) to pass our autoformatter; unfortunately it looks like you didn't check the box that would allow me to make those updates to your PR myself, so are you willing to accept my suggestions so we can get this merged? Thank you!

carljm (changes requested):
Putting it back in your queue for the formatting updates.

Feuermurmel:
> Hi @Feuermurmel, this looks great! It needs a couple minor tweaks (per my comments) to pass our autoformatter

Same issue here: https://github.com/Instagram/MonkeyType/pull/302#issuecomment-1725279372

What's the expectation here? Should I only format the files that I touched and leave the rest as-is, and that will pass the checks?

carljm:
Just format the files within `monkeytype/` directory, and that should also format only the files you touched. Or in this case, you can just accept my suggested edits above and that will also pass the formatter check.

Feuermurmel:
It should be fixed now!

Issue #303, opened 2023-08-26
Keyword-only arguments are not captured in traces


Feuermurmel:
It seems the types of keyword-only arguments are not added to traces:

`__main__.py`:
```python
import foo

foo.uses_kw_only_arg(a=1, b=2)
```

`foo.py`
```python
def uses_kw_only_arg(a, *, b):
    return a + b
```

```text
$ rm monkeytype.sqlite3; pipenv run monkeytype run . && sqlite3 monkeytype.sqlite3 .dump
[...]
INSERT INTO monkeytype_call_traces VALUES('2023-08-26 16:38:40.809497','foo','uses_kw_only_arg','{"a": {"module": "builtins", "qualname": "int"}}','{"module": "builtins", "qualname": "int"}',NULL);
COMMIT;
```

I don't see a fundamental reason for this not working and I think this is just a leftover from older Python versions that did not support keyword-only arguments where `monkeytype.tracing.CallTracer.handle_call()` only looks at `code.co_argcount` and disregards `code.co_kwonlyargcount`.

PR #304, opened 2023-08-26
Capture types of keyword-only arguments

Feuermurmel:
Resolves https://github.com/Instagram/MonkeyType/issues/303.

### DIFF
--- a/monkeytype/tracing.py
+++ b/monkeytype/tracing.py
@@ -215,7 +215,7 @@ class CallTracer:
         if frame in self.traces:
             # resuming a generator; we've already seen this frame
             return
-        arg_names = code.co_varnames[0 : code.co_argcount]
+        arg_names = code.co_varnames[: code.co_argcount + code.co_kwonlyargcount]
         arg_types = {}
         for name in arg_names:
             if name in frame.f_locals:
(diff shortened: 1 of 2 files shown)


monkeytype/tracing.py
```diff
@@ -215,7 +215,7 @@ def handle_call(self, frame: FrameType) -> None:
         if frame in self.traces:
             # resuming a generator; we've already seen this frame
             return
-        arg_names = code.co_varnames[0 : code.co_argcount]
+        arg_names = code.co_varnames[: code.co_argcount + code.co_kwonlyargcount]
```
NotWearingPants, line 218: It seems this can also support `*args` and `**kwargs` by doing
```suggestion
        arg_names = code.co_varnames
```

Feuermurmel: I had that idea too but this alone doesn't work. The code that generates the annotation for the `**kwargs` needs to be adapted as well. The problem is that if a function `foo(**kwargs)` is called as `foo(a=1)`, the correct annotation for `kwargs` would be `int` instead of `dict[str, int]`, which is what happens without additional changes. E.g.:

```python
def uses_kwargs(arg, **kwargs):
    return arg + len(kwargs)

uses_kwargs(1, a=1)
uses_kwargs(1, b=2)
uses_kwargs(1, c='', d='')
```

This will generate the following stub, which doesn't work:

```python
def uses_kwargs(arg: int, **kwargs: Dict[str, Union[str, int]]) -> int: ...
```

carljm:
Thank you for this fix!

Issue #298, opened 2023-08-04
provide a type rewriter to find common base class


dgutson:
Let's assume we have 

file1.py: defines MyEnum1 that extends Enum.
file2.py: likewise MyEnum2
file3.py: defines function F (untyped)

file1 calls F passing an element of MyEnum1, file2 likewise with MyEnum2.

MonkeyType will type F's argument as a Union of MyEnum1 and MyEnum2.
This is not only invalid because it would require a cyclic import (since file3 would need to import file1 and file2 to get the enum definitions), but also the Union explodes when F is a library and is being called from many places.

My proposal is that MonkeyType will find a common base class that can be imported without incurring in a cyclic import.

carljm:
Thanks for the suggestion!

This is a case that falls under MonkeyType's clearly documented minimalist philosophy: it aims to give you accurate and concrete information about the types seen in your program, and let you make the decisions about how to best abstract those into good annotations.

If we added this feature, it would lead to judgment calls that would probably end up requiring new config options. E.g. what happens if there are twenty different subclasses of a common base type and only two of them are ever seen at a particular location? Is the base type still the best annotation, or is the union better?

Or what about cases where the base type is very broad, e.g. some kind of base "object" type in a large type hierarchy? We could special-case the Python `object` type, but users can create similar type hierarchies with similar common bases (e.g. if writing a type-checker in Python!), and may not want all unions reduced to `Object`.

Fundamentally this transformation of the annotation loses potentially useful information, and since we can't be sure it is preferred in all cases, it's better not to do the transformation by default.

This is why MonkeyType has "type rewriters" which allow arbitrary post-transformation of types. The transformation you want can be written as a type rewriter and used in your own projects. I'd even consider merging it into MonkeyType so it's available to all users (though I don't think I would have it be turned on by default.)

carljm:
I'm leaving the issue open in case someone wants to provide a PR implementing this as an optional but built-in type rewriter.

dgutson:
@arieltorti @qequ

carljm:
Fixed with merge of #311

dgutson:
thanks @chrhansk !

PR #311, opened 2023-11-27
Add "most common base" rewriter

chrhansk:
This is an attempted fix for #298: It replaces a union of types by the most common (i.e., most specific) base of all types in the union.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,8 +4,8 @@ Changelog
 main
 ----
 
-* Add rewriter to replace union of derived classes with common
-  base class (not enabled by default)
+* Add ``RewriteMostSpecificCommonBase` type rewriter to replace union of derived classes with common
+  base class (not enabled by default.) Merge of #311, fixes #298. Thanks Christoph Hansknecht.
 
 * Drop support for Python 3.7.
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@ Changelog
 main
 ----
 
+* Add rewriter to replace union of derived classes with common
+  base class (not enabled by default)
+
 * Drop support for Python 3.7.
 
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,7 @@ Changelog
 main
 ----
 
-* Add ``RewriteMostSpecificCommonBase` type rewriter to replace union of derived classes with common
+* Add ``RewriteMostSpecificCommonBase`` type rewriter to replace union of derived classes with common
   base class (not enabled by default.) Merge of #311, fixes #298. Thanks Christoph Hansknecht.
 
 * Drop support for Python 3.7.
--- a/doc/generation.rst
+++ b/doc/generation.rst
@@ -171,6 +171,14 @@ are performed through configured type rewriters.
   type frequently occurs when a method that takes ``List[int]`` also sometimes
   receives the empty list, which will be typed as ``List[Any]``.
 
+.. class:: RewriteMostSpecificCommonBase()
+
+  Rewrites e.g. ``Union[Derived1, Derived2]`` to ``Base``, where
+  ``Derived1`` and ``Derived2`` are both subclasses of a common ``Base``.
+  Unions of arbitrarily many subclasses of a common base class are rewritten
+  as a single class, provided that multiple inherit does not occur in the
+  chain from derived classes to base class.
+
 .. class:: RewriteConfigDict()
 
   Takes a generated type like ``Union[Dict[K, V1], Dict[K, V2]]`` and rewrites
--- a/doc/generation.rst
+++ b/doc/generation.rst
@@ -176,7 +176,7 @@ are performed through configured type rewriters.
   Rewrites e.g. ``Union[Derived1, Derived2]`` to ``Base``, where
   ``Derived1`` and ``Derived2`` are both subclasses of a common ``Base``.
   Unions of arbitrarily many subclasses of a common base class are rewritten
-  as a single class, provided that multiple inherit does not occur in the
+  as a single class, provided that multiple inheritance does not occur in the
   chain from derived classes to base class.
 
 .. class:: RewriteConfigDict()
--- a/monkeytype/typing.py
+++ b/monkeytype/typing.py
@@ -531,17 +531,14 @@ class RewriteMostSpecificCommonBase(TypeRewriter):
             if len(curr_bases) != 1:
                 break
 
-            curr_base = curr_bases[0]
-            curr_klass = curr_base
+            curr_klass = curr_bases[0]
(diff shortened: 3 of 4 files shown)


monkeytype/typing.py
```diff
@@ -501,6 +502,64 @@ def rewrite_Generator(self, typ):
...
+            bases.append(curr_klass)
+
+            curr_bases = curr_klass.__bases__
+
+            if len(curr_bases) != 1:
+                break
+
+            curr_base = curr_bases[0]
+            curr_klass = curr_base
+        return bases[::-1]
```
carljm, line 533: I think what you want here is just `klass.mro()[1::-1]`? That would also handle multiple-inheritance, which you don't handle here.

monkeytype/typing.py
```diff
@@ -501,6 +502,64 @@ def rewrite_Generator(self, typ):
...
+
+    def merge_common_bases(self, first_bases, second_bases):
+        merged_bases = []
+
+        min_len = min(len(first_bases), len(second_bases))
+
+        first_bases = first_bases[:min_len]
+        second_bases = second_bases[:min_len]
```
carljm, line 541: These lines seem unnecessary, as they just duplicate what is already the default behavior of `zip` (if the given iterables are not the same length, "extra" elements in the longer one are ignored)

carljm (changes requested):
Thanks for the PR! The new rewriter in `typing.py` looks reasonable, modulo a couple comments, and the test for it in `test_typing.py` looks good. All the rest of the changes in this PR seems unrelated to the union-rewriter feature, and should be removed from the PR.

The PR should also add documentation for the new rewriter.

monkeytype/typing.py
carljm: Maybe also add some tests covering multiple inheritance cases.

chrhansk:
Thanks for the feedback. I added a test case covering multiple inheritance and fixed up the call to the `zip` function. I also renamed the rewrite to make the intentions more clear. 

Regarding the similarity to `mro()`: The issue in fact revolves around multiple inheritance: The mro states that in case a class is derived from two or more bases, the resolution uses the first of these bases. The rewriter does however enable a partial rewrite even in case of multiple inheritance (see the `test_multiple_bases_nontrivial` test case), which would not be possible based on using mro alone

monkeytype/typing.py
```diff
@@ -501,6 +502,71 @@ def rewrite_Generator(self, typ):
...
+
+    def rewrite_Union(self, union):
+        klasses = union.__args__
+
+        bases = []
+
+        for klass in klasses:
+            base = self._compute_bases(klass)
+            if base is None:
+                return union
```
carljm, line 560: It doesn't seem like there is any scenario in which `_compute_bases` can return `None`? So why include this check?

carljm (changes requested):
Ah, yeah, makes sense that we must not reduce to a common base via multiple inheritance.

A few more comments on the code, and also please run `pipenv run black monkeytype` to fix up the formatting using the autoformatter.

Thank you!

monkeytype/typing.py
chrhansk: You are right, I removed that check.

monkeytype/typing.py
```diff
@@ -501,6 +502,71 @@ def rewrite_Generator(self, typ):
...
+        while curr_klass is not object:
+            bases.append(curr_klass)
+
+            curr_bases = curr_klass.__bases__
+
+            if len(curr_bases) != 1:
+                break
+
+            curr_base = curr_bases[0]
+            curr_klass = curr_base
```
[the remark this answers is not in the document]
chrhansk: I compacted this as well.

monkeytype/typing.py
```diff
@@ -501,6 +502,71 @@ def rewrite_Generator(self, typ):
...
+    def rewrite_Union(self, union):
+        klasses = union.__args__
+
+        bases = []
+
+        for klass in klasses:
+            base = self._compute_bases(klass)
+            if base is None:
+                return union
+            bases.append(base)
```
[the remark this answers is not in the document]
chrhansk: Yep, that was a leftover...

carljm:
Code looks good, thanks for the contribution! Do you feel up to adding some documentation as well? It doesn't need to be extensive, just a mention of the new rewriter and a description of what it does. If not, that's fine, I can add the docs myself, but it may be a couple weeks before I have a chance to do that, and I probably won't merge the PR until docs are ready.

chrhansk:
I added some docstrings to the rewriter itself as well as an entry in the changelog. Is that sufficient?

CHANGES.rst
```diff
@@ -4,6 +4,9 @@ Changelog
 main
 ----
 
+* Add rewriter to replace union of derived classes with common
+  base class (not enabled by default)
```
carljm, line 8: You can also thank yourself here, and mention the issue and PR number :)
```suggestion
* Add ``RewriteMostSpecificCommonBase` type rewriter to replace union of derived classes with common
  base class (not enabled by default.) Merge of #311, fixes #298. Thanks Christoph Hansknecht.
```

carljm (commented):
Sorry, I wasn't clear enough! The changelog entry and docstrings look great, but what I actually meant was the public-facing documentation (published at https://monkeytype.readthedocs.io/en/stable/) which is in the `docs/` subdirectory. Specifically an entry for the new rewriter should be added to the "Type Rewriters" section at the bottom of https://raw.githubusercontent.com/Instagram/MonkeyType/main/doc/generation.rst

chrhansk:
OK, done.

CHANGES.rst
```diff
@@ -4,6 +4,9 @@ Changelog
 main
 ----
 
+* Add ``RewriteMostSpecificCommonBase` type rewriter to replace union of derived classes with common
```
carljm, line 7:
```suggestion
* Add ``RewriteMostSpecificCommonBase`` type rewriter to replace union of derived classes with common
```

doc/generation.rst
```diff
@@ -171,6 +171,14 @@ are performed through configured type rewriters.
   type frequently occurs when a method that takes ``List[int]`` also sometimes
   receives the empty list, which will be typed as ``List[Any]``.
 
+.. class:: RewriteMostSpecificCommonBase()
+
+  Rewrites e.g. ``Union[Derived1, Derived2]`` to ``Base``, where
+  ``Derived1`` and ``Derived2`` are both subclasses of a common ``Base``.
+  Unions of arbitrarily many subclasses of a common base class are rewritten
+  as a single class, provided that multiple inherit does not occur in the
```
carljm, line 179:
```suggestion
  as a single class, provided that multiple inheritance does not occur in the
```

carljm (approved):
Thanks, looks great!

PR #319, opened 2023-12-27
Add reduced sqlite store

chrhansk:
Add a reduced version of the SQLite store optimized for storage efficiency. I will add documentation if you see a chance to merge this.


carljm:
Hi! Appreciate the effort here, and sorry for being slow in getting to this.

I don't really see a need for core MonkeyType to bundle two different SQLite stores. The configurability exists precisely so that such customizations are possible without needing to live in MonkeyType itself. If you think there is broad interest in this beyond your specific situation, I'd encourage you to release it separately on PyPI, and if your intuition is borne out in practice, we can consider a future merger.

PR #320, opened 2024-01-09
Update pipenv dependency

grievejia:
Mainly to move us away from urllib3 2.0.2 as it is marked vulnerable.


carljm:
Ugh, thanks for trying to do this. Pipenv is such a massive headache; ideally we'd probably figure out what to use instead. I am not really sure why that error about monkeytype itself is showing up now and not previously; I wasn't able to replicate that error locally. But in #322 I did manage to get this sorted out; it seems like we don't actually need that editable install of monkeytype in the pipenv deps, so I just removed it.

PR #328, opened 2024-04-15
Update dependencies

pilleye:


pilleye:
Tangentially, I tried updating the test matrix to run on 3.12, but that introduces some breakages in `test_tracing.py`: https://github.com/Instagram/MonkeyType/pull/329

Might be something to look at in the future.

PR #345, opened 2025-01-11
Add Swordk keymap

GigglesnoutWoofle:
Swordkmap

Technical : Added the Swordk_map.py file which contains custom key mappings for the MonkeyType configuration. This setup reassigns keys according to the specified layout so you can practice with it, if you wish.

What it is : "Swordk" is a prototypical keymap loosely based on the philosophies of Qwerty and Colemak designed through the process of trial and error live testing with the goal of being an easier transition from Qwerty.

### DIFF
--- a/monkeytype/Swordk_map.py
+++ b/monkeytype/Swordk_map.py
@@ -1,0 +1,42 @@
+# Swordk_map.py
+from monkeytype.config import DefaultConfig
+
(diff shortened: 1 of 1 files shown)


yangdanny97:
I think the PR is misplaced, this isn't the repository for the "monkeytype.com" typing test.

PR #350, opened 2025-06-14
upgrade upload artifact to v4

aleivag:
acording to https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/ (and our failing test) v3 is not longer supported... upgrading to v4


aleivag:
fails for broken flake8, this fix the upload issue... wil fix flake 8 (move to ruff) in the next commits

PR #351, opened 2025-06-14
use ruff instead of flake8 and black

aleivag:


aleivag:
this fix formatting, now testing is broken... will fix that in the following commit

PR #375, opened 2026-03-13
fix tox config for pipenv and allowlist

zamadye:
hey, saw the issue about tox being outdated. 

updated pipenv to the minimum version that supports python 3.13 (2023.4.29) and added flake8 and black to allowlist_externals since tox 4.24.2 requires it.

let me know if you need anything else!


zamadye:
hmm, having trouble with that. could you clarify?

PR #385, opened 2026-04-07
Fix test_excludes_site_packages for pytest 9

hrnciar:
In pytest 9, `pytest.skip` changed from a plain function to a `_Skip` callable object, so it no longer has a `__code__` attribute. Use `pytest.main` instead, which is still a regular function and serves the same purpose of providing a code object from site-packages.

Assisted-by: Cursor


hrnciar:
The change is simple and I'd need to check with legal whether I am allowed to sign the CLA. Feel free to take the change if you like it.

